PR #7, opened 2018-06-30
chore: add Coveralls reporting

skyeckstrom:


package.json
```diff
@@ -73,6 +72,7 @@
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react-native": "^4.0.0",
     "codecov": "^3.0.2",
```
sdovenor: Does this replace codecov or are we using both?

skyeckstrom: The intent was to replace Codecov with Coveralls but I forgot to remove the old dependency. Updated now.

PR #10, opened 2018-07-02
fix(fscomponents): add type to AddToCart buttonProps

skyeckstrom:


packages/fscomponents/src/components/AddToCart.tsx
```diff
@@ -31,6 +31,7 @@ export interface AddToCartProps {
   swatchesProps?: any;
   stepperProps?: any;
   buttonProps?: any;
```
DShadrick: Did you intend to leave the old any `buttonProps` in here?

skyeckstrom: Definitely didn't. I messed up pulling the changes apart. Fixed.

PR #18, opened 2018-07-03
feat(fscomponents): add textbox field status

sdenaci:
-because Tcomb form native demands that the templates be pure functions, I made stateful versions of all the templates and import them into the functional components.

-to get the styling we want for each template type, and still ensure that users are able to modify the styles by the fieldsStyleConfig prop, I made a modified version of the tcomb native default stylesheet, that has extra classes that are used to make the templates.

-because the active, inactive, and error colors are controlled by state, they are properties on the form component which then get passed into the function that generates the stylesheet and into the stateful template components.


packages/fscomponents/src/components/Form/Form.tsx
```diff
@@ -89,8 +110,17 @@ export class Form extends PureComponent<FormProps> {
       templates
     } = this.props;
 
+    Object.keys(fieldsOptions).forEach(path => {
```
skyeckstrom: We should probably do this and compute the stylesheet in `getDerivedStateFromProps` to avoid re-computing this every render.

packages/fscomponents/src/components/Form/Templates/above/LabelAboveStatefulTextbox.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export default class LabelAboveTextbox extends Component<StatefulTextboxProps,
+StatefulTextboxState> {
+
+  state: StatefulTextboxState = {
+    active: false,
+    validated: false
+  };
+
+  onFocus = () => {
```
skyeckstrom: Looks like we can make these functions `private`

packages/fscomponents/src/components/Form/Templates/above/LabelAboveStatefulTextbox.tsx
```diff
@@ -0,0 +1,108 @@
...
+    });
+  }
+
+  render(): JSX.Element {
+    const {locals} = this.props;
+
+    const prevonBlur = locals.onBlur;
+    const prevOnFocus = locals.onFocus;
+
+    locals.onBlur = () => {
```
skyeckstrom: Can we move this into `getDerivedStateFromProps`?

packages/fscomponents/src/components/Form/Templates/above/LabelAboveStatefulTextbox.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return (
+      <View>
+        <View style={formGroupStyle}>
+          <View>
+            <Text style={[controlLabelStyle, {color}]}>{locals.label}</Text>
+          </View>
+          <View style={textboxViewStyle}>
+            <TextInput
+              accessibilityLabel={locals.label}
+              /* tslint:disable-next-line */
```
pavlemilicevicbb: Hey do you think it is better to define `private input: RefObject<TextInput>;` 
and then initialize it in constructor like this `this.input = React.createRef<TextInput>();` 
and use it like this `ref={this.input}`?

packages/fscomponents/src/components/Form/Templates/above/LabelAboveStatefulTextbox.tsx
bweissbart: I discussed this with Sarah and, after some reading, it looks like this may be a good candidate for memoize: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

She's going to take a stab at doing this then we can revisit.

bweissbart:
@sdenaci - yarn.lock conflicts. If you haven't hit this yet, simply pull down master into your branch, delete your existing yarn.lock file, re-run yarn, then commit the new yarn.lock.

bweissbart:
@skyeckstrom - how's this look now?

bweissbart:
Did a quick sniff test and everything in the demo still seems to be working. The account form is fixed.

PR #25, opened 2018-07-06
docs(fscomponents): reorganize category stories

sdenaci:
Display Category as the main story with the following variants:
-Category Box
-Category Line
-Category Line w/ Flatlist
Update Category Box to have the following knobs:
-View in Grid (1 Column, 2 Columns, 3 Columns)
-w/ Thumbnail (on, off)
Update Category Line to have the following knobs:
-w/ Thumbnail (on, off)
-w/ Accessory (on, off)

To enable on/off knobs for image & accessory, added imageBool property to CategoryBox and CategoryLine components, and acessoryBool prop to CategoryLine. Both props are optional, default to true, and both components still contai necessary checks for image sources before render.


packages/fscomponents/src/components/CategoryLine.tsx
```diff
@@ -19,19 +19,28 @@ export interface CategoryLineProps extends CommerceTypes.Category {
   style?: StyleProp<ViewStyle>;
   titleStyle?: StyleProp<TextStyle>;
   underlayColor?: string;
+  imageBool?: boolean;
```
skyeckstrom: Can we rename these to something more descriptive? Like `shouldShowImage` or just `showImage`?

PR #28, opened 2018-07-06
docs(fscomponents): fix shelf story

sdenaci:
updated CMS values so that default banner would be visible
addded grid of product items


packages/fscomponents/src/components/__stories__/Shelf.story.tsx
```diff
@@ -1,20 +1,79 @@
...
+    },
+    reviews: []
+  },
+  promotions: ['This is a sample promotion!']
+};
+
+const renderProduct = (): JSX.Element => {
+  return (
+    <ProductItemVerticalList
+      {...object('Porduct', testProduct)}
```
skyeckstrom: Porduct should be `Product`

PR #48, opened 2018-07-12
refactor: Make CommerceDataSource an interface

nathan-sankbeil:
CommerceDataSource is being used as an interface rather than an abstract
class, so it should be updated to reflect that.


bweissbart:
testing

PR #93, opened 2018-07-25
docs(fsengage): expand analytics docs

sdenaci:
reordered information, expanded each section with more explanation to give a clearer idea of overall workflow, and added table of contents.


bweissbart:
🥇

PR #105, opened 2018-07-30
fix(fsproductindex): Fix infinite loop of requesting reviews

nathan-sankbeil:
Since the requestReviews method now alters the component's state rather
than triggering an update that results in new props being passed in, the
props/state commerceData in getDerivedStateFromProps were never equal.
This caused the subsequent componentDidUpdate call to always request a
new set of reviews.


nathan-sankbeil:
Hold on reviewing this.

bweissbart:
@nathan-sankbeil - please unblock when you're comfortable with this

nathan-sankbeil:
@bweissbart I think this is good to go now.

PR #121, opened 2018-08-02
feat(pirateship): refactor pdp to allow for inline reloading

bweissbart:
Previously a new screen had to be pushed whenever a user selected a different swatch. This removes the product and review providers and retrieves the data inline, allowing the same screen to be updated with new product data.


nathan-sankbeil:
@bweissbart conflicts, but 👍 otherwise

bweissbart:
I think making PirateShip nicer is a bigger project so holding off on this for now.

PR #130, opened 2018-08-03
fix(pirateship): fixes prods text and shop all cats to dark gray

dfrho:
- Featured Products and Shop All Categories text color set to palette.secondary === grays.eight
- Same with brand text (ie. "Sony")


packages/pirateship/src/screens/Shop.tsx
```diff
@@ -112,8 +113,8 @@ const ShopStyle = StyleSheet.create({
 
 export interface ShopProps
```
pavlemilicevicbb: Since there is no additional props, this can be a type. 
`export type ShopProps = ScreenProps & Pick<........>`

bweissbart:
@dfrho - you don't need to implement pav's suggestion because you didn't modify the types, but you'll have to fix this conflict. Let me know if you need any assistance.

packages/pirateship/src/screens/Shop.tsx
dfrho: add to punch list

packages/pirateship/src/components/PSScreenWrapper.tsx
```diff
@@ -30,13 +30,15 @@ const styles = StyleSheet.create({
...
 
+
 export interface PSScreenWrapperProps {
   style?: StyleProp<ViewStyle>;
 
   hideGlobalBanner?: boolean;
   overrideGlobalBanner?: PSGlobalBannerSlotItem;
 
   needInSafeArea?: boolean;
+  navigator?: Navigator;
```
bweissbart: I don't think this is necessary because we're defining navigator on line 46. If you're getting a tsc error it's probably coming from somewhere else.

dfrho:
just pushed a cleaner version removing the extra navigator props on interface PSScreenWrapperProps.

bweissbart:
@dfrho - conflicts :(

I'd recommend copying your work into a new branch and submitting a new PR.

dfrho:
@bweissbart on it, thanks Brett

PR #131, opened 2018-08-05
fix(pirateship): fixes selected swatch labels

dfrho:
Adds 'Color: [selected color]' and 'Size: [selected size]' to product detail page above swatches


packages/fscomponents/src/components/Swatches.tsx
```diff
@@ -240,7 +242,7 @@ export class Swatches extends Component<SwatchesProps, SwatchesState> {
 
     return (
       <View>
-        {label && this._renderLabel(swatch)}
+        {this._renderLabel(swatch)}
```
nathan-sankbeil: Did we intend to remove the ability to enable/disable swatch labels? If so, can we remove the `label` prop from the `SwatchesProps` interface since it wouldn't be used anymore?

bweissbart: We should keep the code as it was here; you'll need to pass label={true} into the props for this component where it's instantiated (renderSwatches in PSProductDetail.tsx)

PR #138, opened 2018-08-07
fix(fscomponents): fix review stars being cut off

bweissbart:
This fixes the tips of the review stars being cut off, as seen in PirateShip.


nathan-sankbeil:
Reviewing

PR #184, opened 2018-08-29
fix(flagship): Fix possible memory leak and refactor Android code

DanielClayton:
* Remove possible memory leak caused by holding a static reference to an `Activity`
* Optimize code
* Fix spacing
* Optimize imports


bweissbart:
@DanielClayton conflicts

PR #248, opened 2018-10-06
fix(flagship): add targeted device to both Debug and Release config

justin-endler:
Currently targeted device is only injected over the first occurrence of PRODUCT_NAME, which has the side effect of only setting it for the Debug config block (skipping the Release block). If it can be assumed that PRODUCT_NAME will only appear in each "XCBuildConfiguration" of project.pbxrproj, then this is a stable solution. Otherwise, something more complex/thorough should be done.


DavidRhodesBB:
Looks good; tested with PirateShip IOS build.

PR #252, opened 2018-10-10
fix(flagship): use native spawn module to run `react-native link`

VanceHubb:
We've noticed that `react-native link` command in `init` script is not outputing stdout and stderr which are super useful for debugging. Instead, it throws an irrelevant and misguiding AssertionError: `AssertionError [ERR_ASSERTION]: Non-empty queue on spawn exit.`. Looked into the code and found out that we are using `node-suspect` which is a wrapper for native `spawn`. The module focuses on the result of assertions and is not interested in passing stdout and stderr.

Rewrote the code using native node `spawn` after knowing the module is not really needed. Also updated to output stdout and stderr of `react-native link` process so we can locate the real causes of linking failures


bweissbart:
Thanks for taking the time to do this! As of now, `react-native-codepush` has the ability to supply keys via other means than the CLI, so we'll be integrating that feature in the near future and removing custom react-native linking.

PR #273, opened 2018-11-02
feat(fscomponents): added state change logger to Accordion

LamineSadounBB:
needed a way to track whether the accordion has changed state. added prop (function)  that receives stateChanged boolean.


VanceHubb:
@LamineSadounBB the linter is complaining that the format for your commit message is invalid. Try adding a space after the colon, force-pushing that commit, and renaming the pr title accordingly. I can't guarantee this PR will pass after these steps though.

packages/fscomponents/src/components/Accordion.tsx
```diff
@@ -82,6 +82,10 @@ export interface AccordionProps {
    * Function to be invoked when the accordion is opened. Animation will be disabled!
    */
   renderContent?: () => React.ReactNode;
+  /**
+   * Function to be invoked to check the state of the accordion. Takes a callback."
+   */
+  onStateChange?: (stateChanged: boolean) => void;
```
pavlemilicevicbb: callback function does not need to take stateChanged flag, because function will be called only when state is changed. Right?

packages/fscomponents/src/components/Accordion.tsx
```diff
@@ -221,6 +225,13 @@ export class Accordion extends Component<AccordionProps, AccordionState> {
     };
   }
 
+  componentDidUpdate(prevProps: AccordionProps, prevState: AccordionState): void {
+    if (!!this.props.onStateChange && prevState.isOpen !== this.state.isOpen) {
```
pavlemilicevicbb: Function name has to help us determine purpose of the function. It is misleading to call a function `onStateChange` and then just compare `isOpen` flag.
Better name would be something like `onAccordionToggle` and pass `isOpen` flag to a callback.

bweissbart:
Closing due to staleness.

PR #283, opened 2018-11-13
fix(pirateship): replace rn payments dependency with bb fork

bweissbart:
PirateShip is currently failing to run due to linking issues with React Native Payments. This updates Pirateship and fsshopify to use the brandingbrand fork which works and has added functionality to boot.


Cauldrath:
Conflicts.

bweissbart:
This appears to no longer be an issue.

PR #307, opened 2018-12-12
fix(fscomponents): Move Multicarousel opacity and initialized into state

Cauldrath:
These values were not being updated properly when the properties changed, such as changing the images, so opacity was getting stuck set to zero.


jonkoynok:
@bweissbart @skyeckstrom any chance of getting this merged? Appears to be a requirement for https://github.com/brandingbrand/dsg.rn/pull/48

packages/fscomponents/src/components/MultiCarousel/MultiCarousel.tsx
```diff
@@ -84,24 +84,27 @@ export class MultiCarousel<ItemT> extends Component<MultiCarouselProps<ItemT>, M
     this.state = {
       currentIndex: 0,
       containerWidth: 0,
-      itemWidth: 0
+      itemWidth: 0,
+      opacity: new Animated.Value(0),
+      initialized: false
```
skyeckstrom: This is probably the wrong way to handle the animation. We should instead use [`componentDidUpdate `](https://reactjs.org/docs/react-component.html#componentdidupdate) to set the opacity to zero and animate in whenever the image changes. We shouldn't need to keep track of `initialized` at all.

Cauldrath:
Will now animate opacity when the image is changed. It would likely be better to animate when the image is loaded, but that would require more extensive changes to how the image references are handled.

jasonmosleybb:
@bweissbart can you look at this?

bweissbart:
@Cauldrath - commit linter is erroring due to "Merge from upstream" commit message.

Cauldrath:
Squashed commits to fix history.

PR #337, opened 2019-01-23
 fix(pirateship): Allow pirateship to work outside flagship 

Cauldrath:
Builds on: https://github.com/brandingbrand/flagship/pull/326

With these changes, a user should be able to copy the pirateship package from flagship and immediately be able to make it function with minimal changes. I wasn't able to get the rn-cli.config.js file working in both places at this time, but removing the references to parent directories ("../..") and the blacklist should be the only changes needed after copying.

The user should be able to just copy the directory, fix the rn-cli.config.js file, go inside, "yarn install", "yarn run init", then "yarn run ios", "yarn run android", or "yarn run compile-web".


bweissbart:
@Cauldrath - can you please regenerate yarn.lock for this guy

Cauldrath:
Sorry it took so long, but I generated a new yarn.lock file.

jasonmosleybb:
@bweissbart can you look at this again?

Cauldrath:
Merged from upstream/master and updated yarn.lock.

packages/flagship/src/add-keys-android.ts
```diff
@@ -19,7 +19,7 @@ const buildConfig =
   {};
 
 const gradlePropertiesPath =
-  getOption('gradlePropertiesPath') || `../../../../android/gradle.properties`;
+  getOption('gradlePropertiesPath') || `../android/gradle.properties`;
```
bweissbart: This is intended to modify the "ejected" android folder that gets copied to the project root when flagship init is run. Because add-keys-android is typically run after init this will be modifying the android folder inside of flagship instead of the android folder in the root project directory.

packages/flagship/src/lib/link.ts
```diff
@@ -27,7 +27,8 @@ export async function link(configuration: Config): Promise<boolean | object> {
   // Spawn react-native link
   const spawned = spawn('react-native', ['link'], {
     cwd: path.project.path(),
-    shell: os.win
+    shell: os.win,
+    stream: 'stderr'
```
bweissbart: This will surface init errors but if everything is working correctly it won't capture the output from codepush because that's in stdout.

packages/fscomponents/src/components/CMSFeedback.tsx
```diff
@@ -290,8 +290,7 @@ export class CMSFeedback extends Component<CMSFeedbackProps, CMSFeedbackState> {
       feedback: {
         label: FSI18n.string(componentTranslationKeys.form.feedback.label),
         error: FSI18n.string(componentTranslationKeys.form.feedback.error),
-        multiline: true,
-        stylesheet: MultilineStyle
+        multiline: true
```
bweissbart: This should probably remain as-is unless there was a reason for changing this.

packages/fscomponents/src/components/ReviewsSummary.tsx
```diff
@@ -72,7 +72,7 @@ export class ReviewsSummary extends Component<ReviewsSummaryProps> {
         <View style={[S.row, rowStyle]}>
           <Text style={[S.recommendStyle, recommendStyle]}>
             {FSI18n.string(translationKeys.flagship.reviews.recommendCount, {
-              recommendPercent: FSI18n.percent(recommend)
```
bweissbart: Does this need to be removed? The reference project is using a version of this file with FSI18n.percent() still used.

packages/pirateship/package.json
```diff
@@ -5,39 +5,44 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "add-keys-android": "node ../../node_modules/@brandingbrand/flagship/dist/add-keys-android.js",
-    "add-keys-ios": "node ../../node_modules/@brandingbrand/flagship/dist/add-keys-ios-internal.js",
-    "compile-android": "react-native run-android --no-packager",
-    "compile-ios": "react-native run-ios --no-packager --simulator=\"iPhone X\"",
+    "add-keys-android": "node node_modules/@brandingbrand/flagship/dist/add-keys-android.js",
+    "add-keys-ios": "node node_modules/@brandingbrand/flagship/dist/add-keys-ios-internal.js",
+    "compile-android": "(test -d 'android' || yarn init android) && react-native run-android --no-packager",
```
bweissbart: This should be init:android for consistency

packages/pirateship/package.json
```diff
@@ -5,39 +5,44 @@
...
+    "compile-ios": "(test -d 'ios' || yarn init ios) && react-native run-ios --simulator 'iPhone X' --no-packager",
```
bweissbart: init:ios

packages/pirateship/package.json
```diff
@@ -5,39 +5,44 @@
...
-    "@brandingbrand/fsi18n": "2.0.0",
-    "@brandingbrand/fsmockdatasources": "2.0.0",
-    "@brandingbrand/fsnetwork": "2.0.0",
-    "@brandingbrand/fsproductdetail": "2.0.0",
-    "@brandingbrand/fsproductindex": "2.0.0",
-    "@brandingbrand/fssalesforce": "2.0.0",
-    "@brandingbrand/fsshopify": "2.0.0",
-    "@brandingbrand/fsweb": "2.0.0",
+"dependencies": {
+    "@brandingbrand/flagship": "^2.0.0",
```
bweissbart: Lerna handles dependencies inside of flagship so we shouldn't change the version numbers manually for references to other packages in the same repo.

Cauldrath:
Merged in master, pointed at master, and made recommended updates.

packages/pirateship/package.json
Cauldrath: We will still need to change them if we want to copy the package outside of the repo, though. It doesn't seem like it causes any issues while inside the repo, either.

Cauldrath:
Closing because this has gotten too far out of date.

PR #346, opened 2019-02-01
fix: Make scrollToTop function public

pavlemilicevicbb:
So it can be called programatically.


deemaabdallah:
@bweissbart @skyeckstrom  This is the PR we want merged plz!

PR #349, opened 2019-02-05
feat(fscomponents): increase customizability of search bar

BradHabenichtBB:
## Description
adds cancel button style props and a render cancel button prop to the search bar in fscomponents


BradHabenichtBB:
@bweissbart Thanks for restarting. CI tests passed.

PR #351, opened 2019-02-06
feat(fslocator): add default region prop

BradHabenichtBB:
## Description
add default region to fslocator

### DIFF
--- a/packages/fslocator/src/components/LocatorContainer.tsx
+++ b/packages/fslocator/src/components/LocatorContainer.tsx
@@ -64,6 +64,7 @@ export interface PropType {
   renderLocationItemWithBack?: (data: LocationItemData) => JSX.Element;
   renderSearchBar?: (data: SearchBarData) => JSX.Element;
   filterResult?: (data: Location[]) => Location[];
+  defaultRegion?: Region;
   searchEndpoint: string;
   searchRadius?: number;
   resultLimit?: number;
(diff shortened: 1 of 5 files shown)


DavidRhodesBB:
Looks like a typo broke it: 
`src/components/MapView.tsx(100,63): error TS2459: Type 'Readonly<{ children?: ReactNode; }> & Readonly<PropType>' has no property 'defaultRregion' and no string index signature.`

BradHabenichtBB:
@DavidRhodesBB Thanks. Is there a good way to run tsc locally with flagship? When I try it wants to compile the whole project and run tests and all this other junk.

DavidRhodesBB:
@BradHabenichtBB yes, for a faster more efficient build you can leverage the available scripts:

from @nathan-sankbeil :

`yarn lint:ts` <- checks for style issues. runs at project root
`yarn prepare` <- compiles all of the typescript into js, but does not watch for changes. can be run at the project root or in individual packages
`yarn test` <- runs the unit tests. can be run at the project root or in individual packages if they have tests

`yarn ship:run-ios` is comprised of three commands that you could run independently:
`yarn ship:compile-ios` <- builds the native iOS app. if you haven’t changed any native code or added any new modules with a native component, you don’t need to rerun this and can reuse the existing app in the simulator
`yarn ship:run-bundler` <- starts the bundler/server that serves the javascript to the app
`yarn tsc:watch` <- compiles all of the typescript files into js and watches all of the packages for changes. this can be quite resource intensive since it starts a separate node process for each package. if you’re only making changes in a limited number of packages, you can run `yarn tsc:watch` directly in those packages instead

so, let’s say you’re working on some style issues in pirateship and have already built the binary. you could run `yarn ship:run-bundler` at the project root and only run `yarn tsc:watch` in the pirateship directory. that should definitely help stop your computer from bogging down.

PR #353, opened 2019-02-07
fix(flagship): pass leanplum push url to expected launchOptions value

chukohsin:
Move the variable that leanplum sets for the notification url (“_lpx”)  to the default ios variable “UIApplicationLaunchOptionsURLKey” so that it gets passed to the Linking function of React Native

### DIFF
--- a/packages/flagship/ios/FLAGSHIP/AppDelegate.m
+++ b/packages/flagship/ios/FLAGSHIP/AppDelegate.m
@@ -52,6 +52,21 @@
   self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
   self.window.backgroundColor = [UIColor whiteColor];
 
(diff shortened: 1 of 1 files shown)


DavidRhodesBB:
@deemaabdallah can you review?

From @KoHsinChuBB : the issue of this is that when Leanplum sends push notification, it put the url in a place that iOS couldn’t recognize. So we need to set the url to the specific launchOption for iOS. The way to test it will be a bit tricky since you need to send push notification to an iOS app to see whether it can parse correctly.

There is an exact same PR to BBRN.
https://github.com/brandingbrand/BBRN/pull/578/files
Since we have hard time sending push notification to iOS test device in Arcadia, we are not 100% sure that this will fix our issue until we test it in live app.

bweissbart:
Tested and works 👍

PR #365, opened 2019-02-08
fix(fsengage): fix leanplum typings for storybook

bweissbart:
This fixes the LeanplumProvider typings so storybook build works.


DavidRhodesBB:
@bweissbart I can see how this fixes the earlier storybook upgrade build error (cannot have interface named 'LeanPlum') on: https://github.com/brandingbrand/flagship/pull/271

PR #370, opened 2019-02-11
chore(flagship): upgrade react redux to 6.0

DavidRhodesBB:
tested in pirateship


sdenaci:
This looks good, in the future though it would be best to try not to change the ordering of packages in package.json, just for pr clarityreadability

PR #371, opened 2019-02-11
fix(fscomponents): Use onSubmit function for SearchBar button

Cauldrath:
The onSubmit function is not called when the search button is clicked and the search button only calls the onRightBtnPress function, which does not include the value of the search bar. If onRightBtnPress isn't set, onSubmit will be called instead. In the majority of use cases, the functionality of clicking the search button should match submitting the text input, but setting onRightBtnPress will allow it to be overridden for those edge cases.


jasonmosleybb:
@bweissbart can you look at this?

PR #373, opened 2019-02-11
fix(fscommerce): fixes sessiontoken typing

DavidRhodesBB:
attempts to fix typing errors encountered in another PR build:

https://travis-ci.org/brandingbrand/flagship/jobs/458401391


sdenaci:
i am not familiar enough with what this function does or why it was throwing an error in CI. it seems to me it would only ever return `SessionToken | null` but I would like to get @bweissbart 's take on it

bweissbart:
I think this is a good change overall, but I agree with Sarah that `<SessionToken | null>` should suffice here.

PR #381, opened 2019-02-11
fix(selector): added optional accessibility labels

ShaheedShamsiBB:
added optional props accessibility and accessibilityLabel


DavidRhodesBB:
Breaking on CI because we need conventional commits with one of the package names in parentheses '(fscomponents)'. Also the types for accessiblity and accessibilityLabel should be optional.

PR #291, opened 2018-11-20
fix(flagship): Android supportLibVersion downgrade

VanceHubb:
react-native-navigation v1 causing android app crashes:
```
No static method getFont(Landroid/content/Context;ILandroid/util/TypedValue;ILandroid/widget/TextView;)Landroid/graphics/Typeface; in class Landroid/support/v4/content/res/ResourcesCompat; or its super classes (declaration of 'android.support.v4.content.res.ResourcesCompat' appears in base.apk)
```
This is because getFont is deprecated in supportLibVersion 27+. downgrading the version until we upgrade react-native-navigation to v2

PR #407, opened 2019-02-21
fix(flagship): pin android.support libs to 27

bweissbart:
Updates the android.support dependencies in build.gradle to use supportLibVersion constant which is currently set to 27.1.1.

Note: the upgrade to support lib 27 was dictated by RN: https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#tooling-updates


bweissbart:
@DanielClayton

VanceHubb:
@bweissbart @DanielClayton 
We were experiencing Android crashes due to this:  https://github.com/brandingbrand/flagship/pull/291
Theoretically we still need to downgrade support lib to 26 but I haven't tested recently

PR #409, opened 2019-02-21
refactor(fscomponents): accordion some state to props

DavidRhodesBB:
Removing getDerivedStateFromProps(); the state properties it was returning (padding and titleTouchableHighlight) are now pulled directly from props (or the default if not set) in the render function.

Also set background default TouchableHighlight underlayColor to 'transparent' from '#eee' to avoid flashing on press.


DavidRhodesBB:
Depends on:

* [ ]  #411

PR #410, opened 2019-02-21
refactor(fscomponents): actionbar to stateless func component

DavidRhodesBB:
Refactored from stateless class to stateless functional component.


DavidRhodesBB:
Depends on:

* [ ]  #411

PR #419, opened 2019-02-26
refactor(fscomponents): accordion with dev screen

DavidRhodesBB:


DavidRhodesBB:
@sdenaci easy changes - doing now. Yeah I shouldn't have lumped this 'sfc' comment in with the others. We only shaved the state down here and moved some styling to props and defaultProps. Thanks for your help with these!

PR #443, opened 2019-03-04
fix(fscomponents): accessibilitylabel spelling fix

ShaheedShamsiBB:
-fixes an issue with shaheeds ability to spell words


dfrho:
Thanks Shaheed! While we're at it can we add `accessibility=true, accessibilityRole={'text'}` props? That way we can group all child components under the label and accessibility users will know that it's only for reading, not pressing, etc.

PR #456, opened 2019-03-06
fix(fsapp): devMenu path override and web screen wrapper update

Cauldrath:
Allows setting devMenuPath in the config to override the path that the dev menu is displayed on web. Also updates the web screen wrapper to more-closely match the native version's functionality and will suppress the link to the dev menu with the same logic used to determine if the screen should be added to the list.


bweissbart:
@Cauldrath - can you please rebase this? I tried to resolve the conflict in the editor but it's not having any of it.

PR #493, opened 2019-03-16
refactor(fscomponents): sharebutton to stateless function comp

DavidRhodesBB:
used react.memo as it was a PureComponent.

testable in pirateship product detail view, press for share action:

shareButton


dfrho:
replaced by https://github.com/brandingbrand/flagship/pull/512

Issue #492, opened 2019-03-16
react-native-touch-id breaking pirateship on upgrade to 4.0.1


DavidRhodesBB:
error on yarn ship:init

```
> lerna ERR! yarn run prepare exited 1 in '@brandingbrand/pirateship'
> lerna ERR! yarn run prepare stdout:
> yarn run v1.12.3
> $ tsc
> src/components/PSSignInForm.tsx(393,58): error TS2503: Cannot find namespace 'TouchId'.

```
Checking yarn.lock shows upgraded to 4.0.1 with breaking change. Solution: adjust semver of pirateship's react-native-touch-id to "~4.0.0".

bweissbart:
Fixed by https://github.com/brandingbrand/flagship/pull/508

PR #496, opened 2019-03-16
fix(pirateship): corrects semver for rn touch id

dfrho:
blocks auto yarn upgrade of pirateship's **react-native-touch-id** dependency to v 4.0.1 which included breaking changes with their own incorrect semver resulting in `TouchId undefined` errors in **PSSignInForm.tsx**. This Sticks us at 4.0.0 with tilde semver `~4.0.0`.

Fixes: https://github.com/brandingbrand/flagship/issues/492
which is breaking builds/linting


bweissbart:
This was fixed as part of https://github.com/brandingbrand/flagship/pull/508

PR #511, opened 2019-03-18
feat(fscomponents): Custom styling for swatch show more/less

Cauldrath:
Allows setting maxSwatches to 0 or negative to continue showing all swatches even when show less is selected. This also allows specific styling for the container when less or more is selected. This does slightly alter the behavior because the style is applied to a view wrapping just the swatches and not the more/less selection.


Cauldrath:
Added classes that allow adding vender-specific styling to affect the scrollable area's behavior.

Issue #309, opened 2018-12-12
Half Screen Modal Position Prop


jasonmosley:
Add a prop to set the position of the 1/2 screen modal component (top, bottom[default], center).

This will allow the component to have more use cases.

DavidRhodesBB:
submitted under https://github.com/brandingbrand/flagship/pull/517

bweissbart:
Fixed by #846

PR #517, opened 2019-03-20
style(fscomponents): modalhalfscreen vertical position

DavidRhodesBB:
sets up `verticalPosition` props with 'top', 'middle', and 'bottom' as default. Testable in **Development > PSHalfModal** (EmailSignUp, with `verticalPosition={'middle'}`).  

Handles https://github.com/brandingbrand/flagship/issues/309

### DIFF
--- a/packages/fscomponents/src/components/ModalHalfScreen.tsx
+++ b/packages/fscomponents/src/components/ModalHalfScreen.tsx
@@ -1,6 +1,9 @@
 import React, { PureComponent } from 'react';
+import { getStatusBarHeight } from 'react-native-status-bar-height';
+
 import {
   Animated,
+  Dimensions,
   StyleSheet,
   TouchableWithoutFeedback,
   View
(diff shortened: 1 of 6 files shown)


packages/fscomponents/src/components/ModalHalfScreen.tsx
```diff
@@ -19,11 +20,11 @@ const styles = StyleSheet.create({
     bottom: -400
   }
 });
-
 export interface ModalHalfScreenProps {
   visible?: boolean;
   height?: number;
   onRequestClose: () => void;
+  verticalPosition?: string;
```
bweissbart: This should be defined as a enum with values top, middle, bottom: https://www.typescriptlang.org/docs/handbook/enums.html

packages/fscomponents/src/components/ModalHalfScreen.tsx
```diff
@@ -93,16 +94,34 @@ export class ModalHalfScreen extends PureComponent<ModalHalfScreenProps, ModalHa
   }
 
   render(): JSX.Element {
+    let calculatedPosition;
+    const screenHeight = Dimensions.get('window').height;
```
bweissbart: This should be implemented as a Dimensions listener so we don't have to recalculate this value with every render: https://github.com/brandingbrand/flagship/blob/3bfde0951d4a460f26ae2e283ae0996bae8c0e0a/packages/pirateship/src/screens/Category.tsx#L69

DavidRhodesBB: Good stuff, thanks for catching that!  Implemented in updated PR.

packages/fscomponents/src/components/ModalHalfScreen.tsx
DavidRhodesBB: I should have worked harder implementing that - Thanks for the learning on how to use enums @bweissbart.  I also added a small library for calculating the StatusBar height when using 'top' position.

PR #518, opened 2019-03-21
fix(fscomponents): Fixes for IE11 buttons

Cauldrath:
IE11 was collapsing some of the elements in the Button component because they had flex-basis 0


bweissbart:
@Cauldrath - CI is rejecting this guy because scope needs to be the name of a package (missing s in fscomponents)

Cauldrath:
Sorry, I'll just squash the thing, since it will require a force push to change the commit text.

Cauldrath:
Fixed commit history.

PR #520, opened 2019-03-22
feat(fsapp): add ability to pass location into the web router

bweissbart:
The router that we're using for web supports the ability to pass in a location object that overrides the default behavior of using window.location. This allows for the existing web routing system to work with server side rendering as well.

### DIFF
--- a/packages/fsapp/src/types/index.ts
+++ b/packages/fsapp/src/types/index.ts
@@ -68,6 +68,7 @@ export interface AppConfigType {
   popToRootOnTabPressAndroid?: boolean;
   serverSide?: boolean;
   devMenuPath?: string;
+  location?: Location; // Use to provide a server-side location to router in DrawerRouter.web.tsx
 }
 
 export interface NavButton {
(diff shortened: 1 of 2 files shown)


packages/fsapp/src/types/index.ts
```diff
@@ -83,3 +84,10 @@ export interface NavigatorEvent {
   id: string;
   type: string;
 }
+
+export interface Location<T = any> {
+  pathname: string;
+  search: string;
+  hash: string;
+  state: T;
```
Cauldrath: Some of these should probably be optional. If you take the values from window.location, the strings should be defined, but state wouldn't be.

bweissbart: I agree this is wonky, but unfortunately this is a copy of @types/history which is used by react-native-router: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/history/index.d.ts#L28

There were some dependency concerns with importing @types/history into flagship, so I decided to make my own interface since I don't think the history API is going to change anytime soon. :)

We're just going to have to pass empty strings and objects to satisfy the types in the router.

Cauldrath: Makes sense.

PR #529, opened 2019-03-25
feat(fscomponents): make review recommendations configurable

bweissbart:
This updates the "recommended"/"not recommended" tags on ReviewItems configurable. Previously, these tags always displayed if they were made available by the reviews API. There are now four display options: always (default to maintain backwards compatibility), positive only (only display if an item is recommended), negative only (only displayed if an item is not recommended), and never to hide recommendations.


Cauldrath:
Could you make it default to the current behavior so that it's backwards compatible?

bweissbart:
@Cauldrath - I went back and forth on this but per your opinion I went back and made it default to backwards compatible behavior.

PR #532, opened 2019-03-25
chore(fsweb): upgrade cache-loader to 2.0.1

DavidRhodesBB:
The only breaking changes are no more webpack < 4 or Node < 6 so we should be good:
https://github.com/webpack-contrib/cache-loader/releases


DavidRhodesBB:
replaced by #554

PR #558, opened 2019-03-27
feat: change storybook version to 4.1.16

bweissbart:
This appears to resolve an issue in regards to absolute vs relative path for the dll javascript bundle: https://github.com/storybooks/storybook/issues/5055


DavidRhodesBB:
Looks good!

PR #563, opened 2019-03-28
fix(fsapp): Updates to server-side handling of fsapp

Cauldrath:
The old version needed to be used in a way where it created a new instance of FSApp on every request. This way you can create one instance of FSApp and call getApp with the config needed for that request and it will just make a new redux store for each request.

This needs further testing, so I'm opening this right now just so it can be referred to.


bweissbart:
@Cauldrath - is this good to be merged?

Cauldrath:
I haven't heard anything back about how the change is performing or if it has even been deployed, so we should hold off on this until it's been verified as a good change.

PR #564, opened 2019-03-28
chore: refactor ccard validation to remove moment

bweissbart:
This refactors the validation in the CreditCardForm component to eliminate the need for the moment library. Because we were using moment to simply grab the month and date off a MM/YY formatted string it was easy to change. This helps us achieve our goal of removing dependencies to moment because it's a huge library.

**Note:** You cannot test this using Storybook, as the current Storybook component doesn't actually call our custom validation. (As currently set up, you have to manually call validate via a button press and this doesn't exist on the Storybook screen.)

**Note 2:** Moment will still exist in our dependency chain because of the masking library we currently use for credit cards.


vancehu:
Just a suggestion: https://github.com/iamkun/dayjs this library provides minimum date utility functions and is only 6kb unzipped. Maybe we can use this in some cases?

PR #590, opened 2019-04-05
feat(fscomponents): add ability to pass arrow styles for multicarousel

andyschwob:
<MultiCarousel> now accepts 4 new style props to allow styling of the next/prev arrows

`nextArrowContainer`
`nextArrow`
`prevArrowContainer` and
`prevArrow`


packages/fscomponents/src/components/MultiCarousel/MultiCarouselProps.ts
```diff
@@ -18,8 +18,12 @@ export interface MultiCarouselProps<ItemT> {
   itemStyle?: any;
   itemUpdated?: (oldItem: ItemT, newItem: ItemT, index: number, changed: () => void) => void;
   onSlideChange?: (data: SlideChangeEvent) => void;
+  nextArrowContainerStyle?: any;
```
bweissbart: You can make these four types `StyleProp<ViewStyle>` so they're typed correctly coming in. I know we didn't do that with the other style props in this component but might as well start doing it right. :)

PR #601, opened 2019-04-09
chore: tell greenkeeper to ignore react-native updates

bweissbart:
This updates the Greenkeeper configuration to always ignore react-native updates, as we should never be merging in react-native updates without developer oversight. Syntax per the Greenkeeper docs: https://greenkeeper.io/docs.html#monorepo

### DIFF
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -31,5 +31,8 @@
         "packages/pirateship/package.json"
       ]
     }
-  }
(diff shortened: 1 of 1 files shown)


DavidRhodesBB:
@bweissbart Should we also include react and react-native-navigation as in https://github.com/brandingbrand/flagship/pull/368?

PR #603, opened 2019-04-10
feat: android init config

NickBurkhartBB:
## Rationale
During some testing on the related project, I had to add key android:resizeableActivity to the application to prevent split screen mode causing a crash on Android. As I went to make the changes to flagship to allow adding during init process, I checked the repositories of our existing flagship applications and found all places where we are modifying the AndroidManifest.xml and app/build.gradle files in scripts after the init step. I took these scripts as keys are where we could add some template type injection into the Android init process.
I also added config settings to handle version name and codes. These version config settings should allow us to cleanup the manual step for store builds to fix the version numbers that currently exists in many of our projects.

## New Config Values for env configuration
*  **build.additionalDependencies**
list of strings to add to the dependencies block
*  **build.versionName**
string or function that will set version name in gradle.properties. The function will be passed in the current version from package.json
*  **build.versionShortCode**
string or function that will set version short code in gradle.properties. The function will be passed in the current version from package.json
* **build.versionCode**
string or function that will set android.defaultConfig.versionCode in app/build.gradle. The function will be passed in the current version from package.json
* **manifest.activityAttributes**
dictionary of attribute values that will added to the activity element of the AndroidManifest for the MainActivity
* **manifest.additionalElements**
additional xml elements that get added inside the main application element of the AndroidManifest. Useful for adding additional activity elements or services of dependencies
* **manifest.additionalPermissions**
add additional use:permissions elements to the AndroidManifest
* **manifest.applicationAttributes**
dictionary of attribute values that will added to the application element of the AndroidManifest for the MainApplication
* **manifest.urlSchemeHost**
modify deep link host path, some of our apps did not expect url-scheme:/app/* path and just used url-scheme:/* since it works that way on iOS. By allowing this to change we can remove these patch scripts

## Test
Note: these changes should be backwards compatible so they should work on any flagship project
* init flagship project
* ER: should work and compile on Android
* modify the env file for environment your are testing to include android config values specified above
* re-init the project with that environment
* ER: should work and compile and the expected config settings should have been pushed into the Android project


Cauldrath:
Conflicts.

NickBurkhartBB:
@Cauldrath updated

PR #610, opened 2019-04-16
feat(fscomponents): add a11yRole/Label to CategoryLine/SelectableRow

abelemw:


bweissbart:
@abelemw this is deployed as 3.0.1-alpha.75

PR #613, opened 2019-04-16
feat(fscomponents): add accessibility for modalhalfscreen background

chukohsin:
This add accessibilityLabel, accessibilityTraits (iOS) and accessibilityComponentType (Android) to ModalHalfScreen.

### DIFF
--- a/packages/fscomponents/src/components/ModalHalfScreen.tsx
+++ b/packages/fscomponents/src/components/ModalHalfScreen.tsx
@@ -6,6 +6,7 @@ import {
   View
 } from 'react-native';
 import { Modal } from './Modal';
+import { AccessibilityComponentType, AccessibilityTraits } from '../types/Store';
 
 const styles = StyleSheet.create({
   content: {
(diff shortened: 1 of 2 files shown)


packages/fscomponents/src/types/ModalHalfScreen.ts
```diff
@@ -0,0 +1,24 @@
+export type AccessibilityTraits =
```
bweissbart: The linter is complaining because the strings should be in single quotes.

Also, can you please rename this file to something like AccessibilityTypes? I suspect we'll want to use these types in additional places beyond just ModalHalfScreen.

chukohsin: Fixed the linter issue. I didn't change the name since they are the exact names for the properties (accessibilityTraits for iOS and accessibilityComponentType for Android). However, I put the two types in Store.ts so all the components can use the types.

bweissbart:
@chukohsin this is released as 3.0.1-alpha.76

PR #619, opened 2019-04-19
feat(fscomponents): add button properties for carousel

chukohsin:
This changes add button related properties in react-native-swiper to Carousel component. Including showsbutton, nextButton and prevButton.


bweissbart:
@chukohsin this is published as 3.0.1-alpha.78

PR #621, opened 2019-04-19
feat(fscomponents): add completed state accessibility label for Step

chukohsin:
This changes add completed state accessibility label for Step component.

### DIFF
--- a/packages/fscomponents/src/components/Step.tsx
+++ b/packages/fscomponents/src/components/Step.tsx
@@ -11,6 +11,9 @@ import {
   View,
   ViewStyle
 } from 'react-native';
+import FSI18n, { translationKeys } from '@brandingbrand/fsi18n';
+
+const componentTranslationKeys = translationKeys.flagship.step;
 
 const styles = StyleSheet.create({
   completedIcon: {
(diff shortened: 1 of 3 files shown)


packages/fscomponents/src/components/Step.tsx
```diff
@@ -53,6 +53,9 @@ export class Step extends Component<StepProps> {
         <TouchableOpacity
           onPress={this.props.onPress}
           style={[styles.container, this.props.style]}
+          accessibilityTraits='button'
+          accessibilityComponentType='button'
+          accessibilityLabel={`${this.props.title}, Step Completed`}
```
justin-endler: @KoHsinChuBB "Step Completed" should be added to the `fsi18n` package and pulled from there.

packages/fscomponents/src/components/Step.tsx
```diff
@@ -53,6 +53,9 @@ export class Step extends Component<StepProps> {
         <TouchableOpacity
           onPress={this.props.onPress}
           style={[styles.container, this.props.style]}
+          accessibilityTraits='button'
```
justin-endler: `accessibilityTraits` and `accessibilityComponentType` are deprecated for iOS and Android respectively. Can `accessibilityRole` be used instead?

packages/fscomponents/src/components/Step.tsx
chukohsin: @justin-endler  Will fix it, thanks.

packages/fscomponents/src/components/Step.tsx
chukohsin: Yes, we can use it but there will be typescript error if that's okay.

PR #630, opened 2019-04-26
fix(fsengagement): add flag to enable forced resync

andyschwob:
@bweissbart 

Currently if you call `getProfile` as a guest it either gets or creates a profileId based on your app and device.   If you then login with an app we would want to make the api request again with the accountId (email address) to sync it with your profile.  Currently it only checks to see if you have a profileId set and if so it returns the Id without re-syncing.  Passing an optional flag will force the api call to be made that will sync the profile with the account id.


andyschwob:
nevermind, apparently I already put in this pr a while ago

Issue #620, opened 2019-04-19
Switch from sw-precache to Workbox


Cauldrath:
sw-precache has been deprecated. It also uses a different major version of path-to-regexp from fsapp, which can cause conflicts.

https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-sw

This just needs to be changed in these couple places: (I just don't have time to do it myself at this exact moment)
https://github.com/brandingbrand/flagship/blob/ef131c73583156f5954ed8bb988c95f7e8d43387/packages/fsweb/package.json#L32
https://github.com/brandingbrand/flagship/blob/ef131c73583156f5954ed8bb988c95f7e8d43387/packages/fsweb/webpack.config.js#L192

PR #645, opened 2019-05-08
fix(fsweb): Switch to Workbox

Cauldrath:
sw-precache-webpack-plugin has been deprecated and replaced with workbox-webpack-plugin. sw-precache-webpack-plugin was also including an old version of path-to-regexp, which was conflicting with the newer version in some cases.

closes issue #620


bweissbart:
@Cauldrath - conflicts

Cauldrath:
Merged from upstream and regenerated yarn.lock file.

PR #656, opened 2019-05-17
feat(flagship): refactor module scripts and remove node-suspect

bweissbart:
This makes the following changes to the module script system:
- Adds ios and android directories under the modules directory in flagship
- Updated the module system to support two lifecycle functions per module: preLink and postLink
- Updated existing module scripts to replicate the same behavior as before this change: existing ios scripts run as preLink, and existing android scripts run as postLink

The reason for this change is to support updating the react-native-codepush init process. Previously, code-push required its keys to be provided via CLI prompts during init, which required us to implement node-suspect to handle CLI interaction. node-suspect is buggy itself and this solution is hacky.

react-native-codepush recently added support for providing keys in strings.xml (Android) and Info.plist (iOS). However, this requires these files to be modified before react-native link runs, and our implementation of Android module scripts had them run after link. As such, the previously mentioned changes were required in order to allow having an Android script run before rn link.

This shouldn't break Flagship apps using react-native-codepush as it still pulls keys from the same configuration.


Cauldrath:
Testing this with BJs on the hotfix/bopic branch, I'm getting this error that I wasn't getting without this change:
fatal error: module 'RadarSDK' not found

It looks like this is part of the init script:
https://github.com/brandingbrand/bjs.rn/blob/0c6ebb69c256ff0e43d24ea713f07f75a8241f2a/scripts/run_init.sh#L33
https://github.com/brandingbrand/bjs.rn/blob/0c6ebb69c256ff0e43d24ea713f07f75a8241f2a/scripts/radar_io_post_init.js#L70

PR #687, opened 2019-06-20
fix(fscomponents): add accessibility for selectors

Shaheed-Shamsi:
Adds accessibility labels for selector options based on the label name


bweissbart:
@Shaheed-Shamsi - can you please remake this PR in the develop branch? Sorry that this change wasn't communicated.

Shaheed-Shamsi:
@bweissbart I should have closed it, as it is no longer needed. I will close it now

PR #714, opened 2019-07-17
feat(fsengagement): add new carousel option and full screen background

andyschwob:
@bweissbart 

- add new `FullScreenImageCard` for title cards
- add new carousel option for title cards (instead of default list view)

### DIFF
--- a/packages/fsengagement/src/EngagementComp.tsx
+++ b/packages/fsengagement/src/EngagementComp.tsx
@@ -1,6 +1,7 @@
 // tslint:disable:jsx-use-translation-function
 import React, { Component, ComponentClass, Fragment } from 'react';
 import {
+  ActivityIndicator,
   Animated,
   DeviceEventEmitter,
   Dimensions,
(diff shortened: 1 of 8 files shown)


packages/fsengagement/src/EngagementComp.tsx
```diff
@@ -277,10 +316,55 @@ export default function(
...
+    onSnapToItem = (index: number): void => {
+      const pageNum = index + 1;
+      this.setState({
+        pageNum
+      });
+    }
+    // tslint:disable-next-line:cyclomatic-complexity
     renderScrollView(): JSX.Element {
-      const { json: { storyGradient } } = this.props;
+      const { json, json: { storyGradient } } = this.props;
```
bweissbart: Is this a typo?

andyschwob: @bweissbart no, i wanted to have both json and storyGradient as consts, if i just do 
```
const { json: { storyGradient } } = this.props;
```
that doesn't give me json, only storyGradient.

bweissbart: Yep, you're right. New syntax for me!

PR #719, opened 2019-07-22
fix: engagement flicker on scroll of items

NickBurkhartBB:
@deemaabdallah here is the fix for the flickering on the engagement list. There are two parts to the fix.

1.) Making some inbox block components PureComponents
I added shouldComponentUpdate functions to make sure extra re-renders don't occur. I had to use the function instead of PureComponent as each block gets all props so it would still re-render using PureComponent.

2.) adding key to flat list render item
As you scrolled since the keys didn't match it would re-create the element inside the flat list causing the flicker

### DIFF
--- a/packages/fsengagement/src/EngagementComp.tsx
+++ b/packages/fsengagement/src/EngagementComp.tsx
@@ -260,7 +260,7 @@ export default function(
           ...props,
           storyGradient: props.story ? json.storyGradient : null,
           api,
-          key: Math.floor(Math.random() * 1000000)
+          key: this.dataKeyExtractor(item)
         },
         private_blocks && private_blocks.map(this.renderBlock)
       );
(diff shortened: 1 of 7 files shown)


packages/fsengagement/src/EngagementComp.tsx
```diff
@@ -418,5 +419,9 @@ export default function(
         </View>
       );
     }
+
+    dataKeyExtractor = (item: BlockItem): string => {
+      return item.id || item.key || Math.floor(Math.random() * 1000000).toString();
```
NickBurkhartBB: this part is where I was a little unsure of the data structures returned from the API so I tried my best to use the values I felt uniquely identified each part, falling back to the previous random key which might not be the best idea but I left it in

andyschwob: Yeah im pretty sure it's always `key` now, looking at the JSON type i see id is in there too though.  This should be fine.

PR #731, opened 2019-07-31
fix(fsengage): Set up Async Storage dependency

Cauldrath:
Aync Storage is a peer dependency of react-native-permissions 1.2.0. This change will install the dependency and include it in the native projects.

This will allow us to revert pinning the version: https://github.com/brandingbrand/flagship/pull/729


bweissbart:
See https://github.com/brandingbrand/flagship/pull/783 -- we don't need to add any code to the templates because react native link seems to handle the native stuff just fine.

PR #735, opened 2019-08-01
fix: escape android app_name

vancehu:
Unfortunately if the name of an app has xml character entities (For example `Bob's demo`), the android build process will crash. Added escaping logic to fix this


bweissbart:
@vancehu - can you please remake this PR against develop?

PR #736, opened 2019-08-02
chore: remove add-keys scripts in favor of more generic versions

bweissbart:
BREAKING CHANGE: This removes the add-keys-ios and add-keys-android scripts as they required an internal BB workflow to work. The new scripts, add-keys-ios-internal and add-keys-android-internal, can be configured with the environment files with passwords being passed in as environment variables.

For usage instructions, see the new wiki article: https://github.com/brandingbrand/flagship/wiki/Signing-Your-Apps


nathan-sankbeil:
@bweissbart Build failed for Android due to a dependency. The updated lockfile specifies react-native-sensitive-info 5.5, which contains a breaking change (Android X).

PR #739, opened 2019-08-05
feat(fscomponents): add renderModalContent prop

jcarvin:
**Description**
add prop to override modal content

### DIFF
--- a/packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.tsx
+++ b/packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.tsx
@@ -537,7 +537,9 @@ export class ZoomCarousel extends Component<ZoomCarouselProps, ZoomCarouselState
         transparent={true}
         onRequestClose={this.closeZoom}
       >
-        <ZoomImages
+        {this.props.renderModalContent && this.props.renderModalContent(this.closeZoom)}
+        {!this.props.renderModalContent && <View>
+        `<ZoomImages
           gapSizeScaled={this.gapSizeScaled}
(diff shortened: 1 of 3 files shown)


jcarvin:
duplicate of #740

packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.tsx
```diff
@@ -537,7 +537,9 @@ export class ZoomCarousel extends Component<ZoomCarouselProps, ZoomCarouselState
         transparent={true}
         onRequestClose={this.closeZoom}
       >
-        <ZoomImages
+        {this.props.renderModalContent && this.props.renderModalContent(this.closeZoom)}
+        {!this.props.renderModalContent && <View>
+        `<ZoomImages
```
bweissbart: This backtick and the one after `</Animated.View>` shouldn't be there because the entire block is being translated into a string instead of JSX.

jcarvin: My mistake

bweissbart:
@jcarvin - seeing some weird behavior with the zoom carousel in PirateShip (no separate renderModalContent) provided.

Here's a zoomed in image in the v3 branch:

Screenshot 2019-08-05 at 05 59 58 PM

And here's the same zoomed in image with your changes applied:

Screenshot 2019-08-05 at 06 02 29 PM

jcarvin:
@bweissbart It looks like wrapping the contents in a view was causing this behavior.

PR #740, opened 2019-08-05
feat(fscomponents): add renderModalContent prop

jcarvin:
**Description**
add prop to override modal content


jcarvin:
duplicate of #739

packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.tsx
```diff
@@ -537,7 +537,9 @@ export class ZoomCarousel extends Component<ZoomCarouselProps, ZoomCarouselState
         transparent={true}
         onRequestClose={this.closeZoom}
       >
-        <ZoomImages
+        {this.props.renderModalContent && this.props.renderModalContent(this.closeZoom)}
+        {!this.props.renderModalContent && <View>
+        `<ZoomImages
```
bweissbart: This backtick and the one after </Animated.View> shouldn't be there because the entire block is being translated into a string instead of JSX.

jcarvin: My mistake

PR #743, opened 2019-08-05
chore(fsapp): upgrade fsapp rnn to version 2.26.3

chukohsin:
## Note: 
- FSApp Web: used RNN type which should be looked into and removed


bweissbart:
@chukohsin - all of the builds are failing with the error message

```
error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/home/travis/build/brandingbrand/flagship/packages/fsapp/node_modules/react-native-navigation'".
```

PR #775, opened 2019-08-09
fix(fscomponents): fix price/originalPrice comparison

nathan-sankbeil:
The price and originalPrice objects were being compared rather than comparing their values.

BREAKING CHANGE: Under the default render behavior, originalPrice will no longer be displayed unless
it is actually different from price.


bweissbart:
@nathan-sankbeil - needs rebase

nathan-sankbeil:
@bweissbart updated

PR #780, opened 2019-08-17
 chore: upgrade rnn to version 2.26.3

chukohsin:
**This PR includes upgrade the following packages to react-native-navigation v2.26.3:**
- FSApp
- FSShopify
- FSTestproject
- Pirateship


skyeckstrom:
This would be a breaking change for our users. Can you modify your commits so they’re labeled as a breaking change so the version number is incremented correctly?

PR #785, opened 2019-08-21
feat: pin axios to 0.18.0

bweissbart:
v0.19.0 of Axios [contains a regression](https://github.com/axios/axios/issues/2190) that prevents request-specific configuration from being applied. While this has been fixed by PR, a new release has not been made, so this pins Axios to the last good version 0.18.0.


zackseuberling:
howdy y'all just letting you know there was a security flaw in 0.18.0 that was patched, so you may want to pin 0.18.1

PR #793, opened 2019-08-27
feat(fsweb): Enable Progressive Web App

Cauldrath:
Activates features required to build as a progressive web app.

Also switches to Workbox, addressing issue #620


nathan-sankbeil:
@Cauldrath Getting an error when running locally: `Uncaught TypeError: Cannot read property 'Form' of undefined`

Cauldrath:
Added package-lock.json file to fsweb to avoid grabbing new versions automatically and locked some versions to the new values. Also set webpack to 4.40.2 because they fixed the issue in 4.40.0 and added a ship:init-web command for easier debugging.

It looks like the issue was caused by @babel/runtime being installed at version 7.6.0 instead of 7.5.5

PR #796, opened 2019-08-29
fix(flagship): Use React Native Community versions

Cauldrath:
A number of components got moved out of React Native and put into their own repos for React Native 60, including WebView and Geolocation. This switches to those versions.


Cauldrath:
Became part of the RN60 changeset.

PR #824, opened 2019-09-18
fix(pirateship): Search for a free port before starting bundler

Cauldrath:
When running ship:run-ios or ship:run-android, it will search for the first free port starting at 8081 and run against that.

This requires an upgrade to RN 0.60 to work, because there are a number of places in RN where the port is hardcoded and the app will not connect to the bundler:
https://github.com/facebook/react-native/commit/a916dd6632a5e206ab0dbecb89bac8886e1257a5
https://github.com/facebook/react-native/issues/9145


bweissbart:
@Cauldrath - do we want to do anything with this now that we're on RN 61?

Cauldrath:
It should theoretically work without any further changes, but it isn't a very easy thing to test.

Issue #258, opened 2018-10-12
Change how Android project path is handled


bweissbart:
Currently the Android project path is hardcoded to be "com.brandingbrand.reactnative.and.project_name" as specified in getNativeProjectPathAndriod [sic]: https://github.com/brandingbrand/flagship/blob/20a2b2edf30037534403d8f2a5dd509785a246da/packages/flagship/src/lib/path.ts#L171-L181

This value is used in a few places during Android init, such as constructing applicationId in build.gradle.

This is an issue for a couple of reasons:
1. Because Flagship is now open source, application identifiers shouldn't be required to have "brandingbrand" in them
2. We let apps define arbitrary bundle IDs in the `bundleIds.android` environment configuration, so it isn't obvious that the bundle ID actually needs to conform to a strict schema that consists of "com.brandingbrand.reactnative.and." followed by the `name` property in the environment configuration

PR #829, opened 2019-09-24
fix(flagship): Make android namespace match package name

nathan-sankbeil:
The namespace for native android code was always set to
`com.brandingbrand.reactnative.and.{config.name}`, rather than using the defined ID. This lead to some
issues with starting activities and imports.

BREAKING CHANGE: If anything hard codes the android native project path (ie.
`android/app/src/main/java/com/brandingbrand/reactnative/and/project-name`), it will need to be
updated to account for the dynamic package name. Flagship now exports its own modules, so
`flagship.path.android.nativeProjectPath(config)` can be used to get the directory without hard
coding a path.

Will resolve https://github.com/brandingbrand/flagship/issues/258


Cauldrath:
Looks like merging this caused a conflict:
https://github.com/brandingbrand/flagship/pull/603

Cauldrath:
yarn.lock conflicts again.

PR #830, opened 2019-09-24
fix: pin axios to 0.18.x

bweissbart:
Axios 0.19 contains a bug in which custom request parameters are not actually applied to the request.


Cauldrath:
Shouldn't the yarn.lock file also be updated?

bweissbart:
@Cauldrath - good catch. Fixed.

Issue #295, opened 2018-11-26
Update dependencies to fix vulnerability


VanceHubb:
`yarn audit` suggests that we should update those dependencies to fix vulnerability

high:
ws >= 3.3.1

medium:
morgan  >=1.9.1

low:
merge >=1.2.1
lodash >= 4.17.5

PR #831, opened 2019-09-24
fix(flagship): Upgrade vulnerable modules

Cauldrath:
Upgrade typedoc, typedoc-plugin-monorepo, and commitlint to newer versions so lodash and marked no longer are flagged as having high and moderate vulnerabilities in yarn audit.

Documentation generation has been broken since upgrading Typescript, so it fails with and without this change.

Related issue (not resolved due to 2088 low priority vulnerabilities to be resolved): https://github.com/brandingbrand/flagship/issues/295


bweissbart:
This was fixed with https://github.com/brandingbrand/flagship/commit/f6b6c2137df34ffb88eb9bba6c7571007324f6ed. Apologies for missing your PR for this one!

PR #839, opened 2019-09-30
fix(pirateship): Disable on press when scrolling on grid

MauricioPerezBB:
Test:
This ticket fixes web.
On any brower go to a product list. Scroll down and it should not take you to any product detail until you stop scrolling.


Cauldrath:
Relevant issues in react-native-web:
https://github.com/necolas/react-native-web/issues/1219
https://github.com/necolas/react-native-web/issues/731

This change only addresses the issue specifically in PirateShip Product Index and will not fix it for any existing apps or any other sections of the site. The issue seems to be specific to touchables inside of ScrollViews.

PR #840, opened 2019-10-01
feat(flagship): support background modes

MauricioPerezBB:
Improvement for the iOS configuration project

Update the info.plist so it accept background modes capabilities

### DIFF
--- a/packages/flagship/src/lib/ios.ts
+++ b/packages/flagship/src/lib/ios.ts
@@ -287,7 +287,7 @@ export function backgroundModes(configuration: Config): void {
     '<key>UIRequiredDeviceCapabilities</key>',
     `<key>UIBackgroundModes</key>
       <array>
-      ${configuration.UIBackgroundModes.forEach(mode => {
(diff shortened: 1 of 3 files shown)


packages/flagship/src/lib/ios.ts
```diff
@@ -259,13 +259,42 @@ export function usageDescription(configuration: Config): void {
...
+  );
+
+  const infoPlist = path.ios.infoPlistPath(configuration);
+
+  fs.update(
+    infoPlist,
+    '<key>UIRequiredDeviceCapabilities</key>',
+    `<key>UIBackgroundModes</key>
+      <array>
+      ${configuration.UIBackgroundModes.forEach(mode => {
```
Cauldrath: The `forEach` here needs to be `map`. As is, this just puts "undefined" in the file.

MauricioPerezBB: @Cauldrath you're right, PR updated.

PR #846, opened 2019-10-03
feat: add position property for half modal component

sentiurin:
### Description:
#309

This PR was closed because it was old and had conflicts.
#517

Let's take what David did and create a fresh PR.

### DIFF
--- a/packages/fscomponents/src/components/ModalHalfScreen.tsx
+++ b/packages/fscomponents/src/components/ModalHalfScreen.tsx
@@ -1,6 +1,8 @@
 import React, { PureComponent } from 'react';
 import {
   Animated,
+  Dimensions,
+  NativeModules,
   StyleSheet,
   TouchableWithoutFeedback,
   View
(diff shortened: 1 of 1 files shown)


packages/fscomponents/src/components/ModalHalfScreen.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { PureComponent } from 'react';
 import {
   Animated,
+  Dimensions,
```
Cauldrath: We should avoid using Dimensions any time before componentDidMount is called, if possible, because it can mess up server-side rendering, where screen dimensions are not available.

sentiurin: @Cauldrath So, as i understand we can use Dimensions after componentDidMount called?

Cauldrath: Correct. So you could set the height in state to a default value in the constructor and then grab it from Dimensions in componentDidMount. You should also check for updates to it, in case the user resizes a browser window or rotates a mobile/tablet device. Here's an example of setting up the Dimension listener:
https://github.com/brandingbrand/cvs.fs/blob/develop/packages/otchs-web/src/screens/Shop.tsx#L50

If you could get it to work with percentage-based positioning, that would avoid all that, though.

sentiurin: Good think about percent positioning. I’ll check for state animation work with it (coz not sure), and will create one more commit.

packages/fscomponents/src/components/ModalHalfScreen.tsx
```diff
@@ -21,33 +23,96 @@ const styles = StyleSheet.create({
...
 }
 
 export class ModalHalfScreen extends PureComponent<ModalHalfScreenProps, ModalHalfScreenState> {
   constructor(props: ModalHalfScreenProps) {
     super(props);
     this.state = {
       visible: false,
-      contentOffset: new Animated.Value(0)
+      contentOffset: new Animated.Value(0),
+      height: 0,
```
Cauldrath: You should be able to set this to the height prop. If that is being set incorrectly, it's a problem with the parent component.

Issue #334, opened 2019-01-17
Add archive/assembleRelease checks to make sure dev menu is disabled


bweissbart:
The dev menu is enabled or disabled based on the `disableDevFeature` boolean in the environment variable. Upon init, the value of this environment variable is set as a native constant, which is then checked when the app runs to determine whether the dev menu should be shown: https://github.com/brandingbrand/flagship/blob/22984e378e9d73f0c1c91db3e75fc73b303059f7/packages/fsapp/src/components/screenWrapper.tsx#L69

We should add checks to the archive (iOS) and assembleRelease (Android) steps to build store releases that verifies that the dev menu is disabled, and throws an error if that's not the case.

PR #852, opened 2019-10-04
fix: Add archive/assembleRelease checks to make sure dev menu is disabled

alexandrvlasov:
#334 

The dev menu is enabled or disabled based on the disableDevFeature boolean in the environment variable. Upon init, the value of this environment variable is set as a native constant, which is then checked when the app runs to determine whether the dev menu should be shown:

https://github.com/brandingbrand/flagship/blob/22984e378e9d73f0c1c91db3e75fc73b303059f7/packages/fsapp/src/components/screenWrapper.tsx#L69

We should add checks to the archive (iOS) and assembleRelease (Android) steps to build store releases that verifies that the dev menu is disabled, and throws an error if that's not the case.


bweissbart:
We distribute UAT release builds that do have the dev menu intentionally enabled, so we can't put this in. I think the only correct course of action is to just make sure we validate our builds before we submit them to the app stores.

Issue #340, opened 2019-01-25
Add 'Marquis' feature to Button with title dynamically reflecting CTA state


DavidRhodesBB:
For example, when successfully adding to cart, the button title shows 'Added to Cart' for 2 seconds then flips back to 'Add to Cart'.  Other CTA states could be 'Send', 'Sending', 'Sent', for example.

bweissbart:
Fixed with #853

PR #853, opened 2019-10-04
fix: Add 'Marquis' feature to Button with title dynamically reflecting CTA state

olksns:
### Description
Reply on Issue/Reaquest https://github.com/brandingbrand/flagship/issues/340;

Now the button have two additional not required properties: 
`dynamicTitleStates: string[];`
This property require not empty string array that contains all title states you need;

`selectedTitleState: number;`
This property is responsible for the selected dynamicTitle index;

If the button title has no above mentioned props or props are invalid Button component should return the default title property;

### Test
1. Run test project on ios/android;
2. Click on any Item and open PDP page;
3. Click on Add to Cart button;
4. Observe;

_Expected result:_ 
Button should change the state to Added to Cart and then change state back to Add to Cart;


packages/fscomponents/src/components/Button.tsx
```diff
@@ -147,4 +157,26 @@ export class Button extends PureComponent<ButtonProps, ButtonState> {
...
+      selectedTitleState,
+      dynamicTitleStates
+     } = this.props;
+
+    if (
+      (!selectedTitleState && selectedTitleState !== 0) ||
+      !dynamicTitleStates ||
+      (
+        selectedTitleState &&
+        dynamicTitleStates &&
```
Cauldrath: You shouldn't need these two checks because it won't reach this part if these are falsey (except for if selectedTitleState is zero, but that would also be fine). Also, selectedTitleState should only be a number or undefined, so you should be able to just check to see if it is undefined.

olksns: @Cauldrath done

Issue #562, opened 2019-03-28
running storybook warns of not declaring core-js version


DavidRhodesBB:
Storybook upgrade has some soft warnings during the build for upcoming version requirements: 

```
WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3

```

Stories are housed in the **fscomponents** package.

bweissbart:
Fixed with #850

PR #856, opened 2019-10-04
fix(flagship): Upgrade storybook to 5

Cauldrath:
Fixes https://github.com/brandingbrand/flagship/issues/562
Caused by https://github.com/storybookjs/storybook/issues/6261
Upgraded to 5.0.11 because 5.1 upgrades core-js to 3, which causes conflicts with @babel/polyfill, which is still on 2 on the latest version.


Cauldrath:
Closing because this has been moved into Flagship 9, and is already included in the RN61 PR.

Issue #144, opened 2018-08-08
Development version number should be fixed to the bottom of viewport for web


nathan-sankbeil:
Currently shows after the last element on the page, so it's annoying to access on pages with a lot of content.

![screen shot 2018-08-08 at 1 17 48 pm](https://user-images.githubusercontent.com/3847536/43853294-b9ce21f0-9b0d-11e8-80dd-95c213410c7f.png)

bweissbart:
Fixed by #857

PR #857, opened 2019-10-07
fix: Development version number should be fixed to the bottom of view…

Bodrus:
Description:
#144

Currently shows after the last element on the page, so it's annoying to access on pages with a lot of content.

Changes:
Fix: Development version number is fixed to the bottom of viewport for web, now


packages/fsapp/src/components/screenWrapper.web.tsx
```diff
@@ -7,13 +7,15 @@ import qs from 'qs';
 import pushRoute from '../lib/push-route';
 import { AppConfigType, DrawerConfig } from '../types';
 
-const styles = StyleSheet.create({
+const StyleSheetCreate: any = StyleSheet.create;
```
Cauldrath: I'm not sure why you made a separate variable for this, especially because it's only called once.

Cauldrath: Ah, makes sense. I usually just do a ts-ignore for the specific line, because this shuts off any type checking for the entire object, but this should be fine.

PR #858, opened 2019-10-07
fix: fix infinite re-render bug when using renderSyndicatedIndicator prop

jcarvin:
fix: fix infinite re-render bug when using renderSyndicatedIndicator prop

### DIFF
--- a/packages/fscomponents/src/components/ReviewItem.tsx
+++ b/packages/fscomponents/src/components/ReviewItem.tsx
@@ -90,13 +90,6 @@ export class ReviewItem extends Component<ReviewItemProps, ReviewItemState> {
     }
   }
 
-  getImageSizeSuccess = (w: number, h: number) => {
-    this.setState({
-      syndicatedImageHeight: h,
-      syndicatedImageWidth: w
-    });
-  }
-
   renderSyndicatedIndicator = (): JSX.Element | undefined => {
     if (this.props.syndicationSource && this.props.syndicationSource.LogoImageUrl) {
       if (this.props.renderSyndicatedIndicator) {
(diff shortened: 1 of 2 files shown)


packages/fscomponents/src/components/ReviewItem.tsx
```diff
@@ -97,39 +97,46 @@ export class ReviewItem extends Component<ReviewItemProps, ReviewItemState> {
...
+              site: this.props.syndicationSource.Name
+            })}
+          </Text>
+        </View>
+      );
+    } else if (
+      this.props.renderSyndicatedIndicator &&
+      this.props.syndicationSource &&
+      this.props.syndicationSource.LogoImageUrl) {
+      return this.props.renderSyndicatedIndicator(this.props.syndicationSource);
```
Cauldrath: Making the if checks like this should make it a little cleaner and require fewer checks:
```
if (this.props.syndicationSource && this.props.syndicationSource.LogoImageUrl) {
  if (this.props.renderSyndicatedIndicator) {
    return this.props.renderSyndicatedIndicator(this.props.syndicationSource);
  } else {
    return (big block of jsx);
  }
} else {
  return;
}
```

You might even be able to cut the last return.

packages/fscomponents/src/components/ReviewItem.tsx
```diff
@@ -98,38 +99,16 @@ export class ReviewItem extends Component<ReviewItemProps, ReviewItemState> {
 
   renderSyndicatedIndicator = (): JSX.Element | undefined => {
     if (this.props.syndicationSource && this.props.syndicationSource.LogoImageUrl) {
-
-      Image.getSize(
-        this.props.syndicationSource.LogoImageUrl,
-        this.getImageSizeSuccess,
```
Cauldrath: Since you are removing the reference here, you should be able to remove the function, too.

jcarvin: Good Catch. Removed.

PR #859, opened 2019-10-07
feat(fsengagement): expose new inbox route that sets attributes

andyschwob:
This exposes a new method `getInboxMessages` which uses the new route that set user attributes and fetches the inbox in the same request.

### DIFF
--- a/packages/fsengagement/src/EngagementService.ts
+++ b/packages/fsengagement/src/EngagementService.ts
@@ -2,6 +2,7 @@ import FCM, { FCMEvent } from 'react-native-fcm';
 import AsyncStorage from '@react-native-community/async-storage';
 import FSNetwork from '@brandingbrand/fsnetwork';
 import DeviceInfo from 'react-native-device-info';
+import moment from 'moment';
 import {
   EngagementMessage,
   EngagementProfile,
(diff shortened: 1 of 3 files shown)


packages/fsengagement/src/EngagementService.ts
```diff
@@ -237,6 +241,57 @@ export class EngagementService {
...
+          message: JSON.parse(data.message),
+          title: data.title,
+          inbox: data.inbox,
+          attributes: data.attributes
+        };
+      }))
+      .then((messages: EngagementMessage[]) => {
+        this.messages = messages;
+        this.messageCache = +new Date();
+        AsyncStorage.setItem('LAST_ENGAGEMENT_FETCH', moment.utc(new Date()).valueOf().toString())
```
Cauldrath: Because moment is such a large library, could you switch to using just the base Date functions or a smaller library to format this?

andyschwob: Sure, this code is already merged into the rnn2 branch, should we change it there as well?

Cauldrath: Yeah, I noticed moment was added there when I was doing a merge, but wasn't sure what it was for. If you could make the same change there, I would appreciate it.

Cauldrath: Sorry, I forgot that we should be able to just merge develop into that branch and it should make the same change. I hope you haven't put a bunch of work into that.

PR #865, opened 2019-10-09
fix(flagship): add support for new pods

MauricioPerezBB:
This PR add support to add new dependencies to the configuration object and modify the podfile in the init script.

### DIFF
--- a/packages/flagship/src/types.ts
+++ b/packages/flagship/src/types.ts
@@ -152,7 +152,8 @@ export interface IOSConfig {
 }
 
 export interface PodsConfig {
-  sources: string[];
+  sources?: string[];
+  newPods?: string[];
 }
 
 export interface NPMPackageConfig {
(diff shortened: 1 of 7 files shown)


packages/flagship/src/types.ts
```diff
@@ -153,6 +153,7 @@ export interface IOSConfig {
 
 export interface PodsConfig {
   sources: string[];
+  newPods: string[];
```
Cauldrath: You should make this optional so it's backwards compatible, including adding a check in `packages/flagship/src/commands/init.ts` to see if it's undefined.

Cauldrath: Actually, you should make sources optional, too, in case there's ever a need to add new pods without adding sources.

MauricioPerezBB: @Cauldrath PR updated, but in fact there's no need for a sources optional, we're already checking if sources exist inside the function in packages/flagship/src/lib/cocoapods.ts   **if (sources && sources.length > 0) {**

MauricioPerezBB: @Cauldrath  Let me do the same with sources then, so we have the same convention, and squash the commits to 1

Cauldrath: The Typescript object should still allow setting it to undefined.

MauricioPerezBB: @Cauldrath PR updated.

PR #868, opened 2019-10-09
fix(fscomponents): Fix use of empty star style for reviews

Cauldrath:
Was passing the entire style into color


Cauldrath:
Fixed the style order.

bweissbart:
I believe that the swap needs to occur on the second changed instance as well.

PR #873, opened 2019-10-15
fix(fsengagement): refactor animated story modal and fix deeplinking

andyschwob:
### Changes

- change animated story from an overlay to a modal
- for deeplinks from engagement, add query parameters to signify deeplink is from a story and also to contain componentId
- with update to react-native-device-info there are a couple `DeviceInfo` changes needed because of api change
- remove a few logEvent calls that aren't needed anymore


andyschwob:
@bweissbart definitely double check my commit to make sure it's ok, but it looks like the reverse-jetifier worked and passed the android-build script

Issue #241, opened 2018-10-05
Replace webpack-serve with webpack-dev-server


nathan-sankbeil:
[webpack-serve](https://github.com/webpack-contrib/webpack-serve) is now deprecated and should be replaced by [webpack-dev-server](https://github.com/webpack/webpack-dev-server)

PR #874, opened 2019-10-17
build: upgrade deprecated webpack dev server

olexandor:
Resolve issue #241

### Description:
Changed deprecated webpack server to new webpack dev server.

### DIFF
--- a/packages/fsweb/webpack.config.js
+++ b/packages/fsweb/webpack.config.js
@@ -247,6 +247,7 @@ module.exports = function(env, options) {
     (!options || !options.json) && console.log('Webpacking for Development');
     globalConfig.devServer = {
       contentBase: path.join(__dirname, 'dev-server'),
+      historyApiFallback: true,
       port: 8080
     };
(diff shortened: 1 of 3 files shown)


packages/fsweb/webpack.config.js
```diff
@@ -233,12 +231,10 @@ module.exports = function(env, options) {
     ]);
   } else {
     (!options || !options.json) && console.log('Webpacking for Development');
-    globalConfig.serve = {
-      content: "./dev-server",
-      add: (app, middleware, options) => {
-        app.use(convert(history()));
-      }
```
Cauldrath: This made it so that you could visit paths besides the root one and have the app still show up. To do the same with webpack-dev-server, you'll need to add `historyApiFallback: true`

Also, please update the yarn.lock file.

olexandor: Option historyApiFallback added to webpack-dev-server configuration. yarn.lock file updated.

PR #881, opened 2019-10-24
feat(fsshopify): account data source implementation

MauricioPerezBB:
This PR implements methods for account data management in Shopify and expose a data source API


jasonmosley:
@bweissbart This looks good to me. Can we get this merged?

packages/fsshopify/src/mixins/ShopifyAccountDataSource.ts
```diff
@@ -0,0 +1,364 @@
...
+              provinceCode
+              zip
+            }
+          }
+        }
+      }
+    }`, {
+      customerAccessToken
+    });
+    const addresses = response.customer.addresses.edges.map((address: any) => {
```
Cauldrath: We should make sure this exists before accessing it. Same with defaultAddress below.

packages/fsshopify/src/normalizers/blogPosts.ts
```diff
@@ -0,0 +1,25 @@
...
+export function blogPosts(blogData: any): Article {
+  return {
+    author: blogData.node.authorV2.name,
+    image: blogData.node.image.originalSrc,
+    description: blogData.node.contentHtml,
+    excerpt: blogData.node.excerpt,
+    title: blogData.node.title,
+    publishedAt: blogData.node.publishedAt,
+    url: blogData.node.url,
+    id: blogData.node.id,
```
Cauldrath: Should also add checks here.

jasonmosley: I am not sure how many people will want to consume the blog feed so maybe we should remove this functionality all together and really just focus in ecom/user related features.

MauricioPerezBB: @Cauldrath @jasonmosley PR updated.

PR #882, opened 2019-10-24
fix: redundant summary to statistics

ooleksolo:
### Description

To avoid redundant review summary or statistics, the { ... statistics: ReviewStatistics } property was removed from project as it extend the ReviewSummary, has no required data. Also ReviewStatistics has very similar usage as the ReviewSummary. 

### Steps to test
All components that where using the ReviewStatistic and/or ReviewSummary should work as before.


bweissbart:
Closing due to large conflicts.

Issue #452, opened 2019-03-05
Metro bundler should not watch src folder


VanceHubb:
Metro bundler is currently watching src folder and will trigger a live-reloading event when there is a ts file being saved. However, we are expecting the event to be only fired after `tsc` watch compiled `src` to `dist`. This conflict causes inconsistency between `src` and `dist` until the user saves the file again.

Need to exclude the `src` folder or only watch `dist` folder, whichever is more appropriate

vancehu:
since metro bundler supports typescript source files,  all we need to do is (pirateship and actual projects):
1. not emitting `dist` folder in tsconfig 
2. set entry point to import `src` instead of `dist` folder
And the issue above can be fixed

PR #884, opened 2019-10-25
fix: Metro bundler should not watch src folder

alexandrvlasov:
#452 

## Description
Metro bundler is currently watching src folder and will trigger a live-reloading event when there is a ts file being saved. However, we are expecting the event to be only fired after tsc watch compiled src to dist. This conflict causes inconsistency between src and dist until the user saves the file again.

Need to exclude the src folder or only watch dist folder, whichever is more appropriate


alexandrvlasov:
@bweissbart I changed.

bweissbart:
I'm unable to run the project with these changes. Since this only affects local dev I don't think it's worth spending additional time to investigate.

Issue #344, opened 2019-01-31
Add non-FlatList Grid Component


bweissbart:
https://github.com/brandingbrand/flagship/blob/master/packages/fscomponents/src/components/Grid.tsx

The Grid component, as currently implemented above, utilizes a FlatList. This was most likely done because it was intended to be used in interfaces that are expected to scroll, such as Product and Category index pages.

However, projects are using Grid as a table view, and in most cases don't expect it to be using a FlatList behind the scenes. We should make a component that simply displays table data separated by optional grid lines that doesn't implement a FlatList or ScrollView on its own.

OlgaKutovaya:
@bweissbart we have already fixed this issue #885

bweissbart:
Fixed with #885

PR #885, opened 2019-10-25
fix: non-FlatList Grid Component

OlhaKutova:
## Description

Fix for #344 

The Grid component currently don't utilizes a FlatList behind the scenes. The Grid component now simply displays table data separated by optional grid lines that doesn't implement a FlatList or ScrollView on its own.


Cauldrath:
This removes the back-to-top button functionality. It looks like none of our sites are using it, though, and the logic should probably be applied elsewhere but it does mean that this will wind up a breaking change. It might be better to make this to the feature/rn-60 branch so it is grouped with other breaking changes. (Not requesting any changes at this time, just making a note here.)

bweissbart:
Closing because we've gone a different direction with components.

PR #886, opened 2019-10-28
chore: Update react-native-web

alexandrvlasov:
#209
The devDependency react-native-web was updated from 0.11.1 to 0.11.6


Cauldrath:
Updated to 0.11.7 in https://github.com/brandingbrand/flagship/pull/915

PR #890, opened 2019-10-28
chore: Update @types/react-router-dom

alexandrvlasov:
#79 

An in-range update of react-native-sensitive-info is breaking the build


bweissbart:
This was addressed with a Greenkeeper PR.

PR #891, opened 2019-10-28
update: update @types/react dependency

Bodrus:
**Ticket:**
Update @types/react dependency

**Mocks:**
https://github.com/brandingbrand/flagship/issues/772

Fix: The devDependency @types/react was updated from to 16.9.11.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "@types/memoize-one": "^4.1.1",
     "@types/pluralize": "^0.0.29",
     "@types/qs": "^6.5.1",
-    "@types/react": "^16.6.3",
+    "@types/react": "^16.8.3",
     "@types/react-dom": "^16.9.3",
     "@types/react-native": "^0.57.30",
     "@types/react-native-i18n": "^2.0.0",


yarn.lock
```diff
@@ -3130,14 +3130,22 @@
...
+"@types/react@*", "@types/react@^16.8.8":
   version "16.8.24"
   resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.24.tgz#8d1ea1fcbfa214220da3d3c04e506f1077b0deac"
   integrity sha512-VpFHUoD37YNY2+lr/+c7qL/tZsIU/bKuskUF3tmGUArbxIcQdb5j3zvo4cuuzu2A6UaVmVn7sJ4PgWYNFEBGzg==
   dependencies:
     "@types/prop-types" "*"
     csstype "^2.2.0"
 
+"@types/react@^16.9.11":
+  version "16.9.11"
```
Cauldrath: Having two versions of the same module can cause problems and this doesn't fully upgrade to the new version. All of the above dependencies for `@types/react` can match 16.9.11, so you should consolidate them all to use this one.

Cauldrath: Now this PR doesn't affect anything, because it's already loading 16.8.3. This also doesn't address the issue, which was for upgrading to at least 16.8.25.

bweissbart:
I think we should only update our types when we explicitly upgrade React or React Native in Flagship. I made a change to our Greenkeeper config to ignore this package.

PR #892, opened 2019-10-28
fix: Fix tests with date format helpers.

olksns:
### Description
Fix tests with date format helpers.

### Steps to test
1. yarn test
2. all tests should be passed


bweissbart:
Our tests have been passing so I don't think this is an issue.

PR #902, opened 2019-10-29
chore(fsweb): update babel-plugin-react-native-web

sentiurin:
Update babel-pluigin-react-native-web for #73


Cauldrath:
This doesn't actually update to a newer version, as the version in the yarn.lock file is the same. https://github.com/brandingbrand/flagship/pull/915 has upgraded to 0.11.7

PR #903, opened 2019-10-29
chore: update @types/storybook dependency

Bodrus:
**Ticket:**
Update @types/storybook dependency

## Description
The devDependency @types/storybook__addon-actions was updated from 3.4.2 to 3.4.3

Issue: https://github.com/brandingbrand/flagship/issues/682

Updating dependencies require understanding / researching what is changing in the update and ensuring that current projects won't break. If the update breaks our exitsing projects, then we should not update

** Done**
@types/storybook__addon-actions was updated  to 3.4.3

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3168,7 +3168,7 @@
   resolved "https://registry.yarnpkg.com/@types/stack-utils/-/stack-utils-1.0.1.tgz#0a851d3bd96498fa25c33ab7278ed3bd65f06c3e"
   integrity sha512-l42BggppR6zLmpfU6fq9HEa2oGPEI8yrSPL3GITjfRInppYFahObbIQOQK3UGxEnyQpltZLaPe75046NOZQikw==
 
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -3168,7 +3168,7 @@
   resolved "https://registry.yarnpkg.com/@types/stack-utils/-/stack-utils-1.0.1.tgz#0a851d3bd96498fa25c33ab7278ed3bd65f06c3e"
   integrity sha512-l42BggppR6zLmpfU6fq9HEa2oGPEI8yrSPL3GITjfRInppYFahObbIQOQK3UGxEnyQpltZLaPe75046NOZQikw==
 
-"@types/storybook__addon-actions@^3.0.3":
+"@types/storybook__addon-actions@^3.4.3":
   version "3.4.3"
```
Cauldrath: 3.4.3 was already being used, so this PR doesn't change anything.

bweissbart:
Closing per comment.

PR #904, opened 2019-10-29
chore: Update react-native-sensitive-info

alexandrvlasov:
#97 
An in-range update of react-native-sensitive-info is breaking the build

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -13053,7 +13053,7 @@ react-native-safe-area@^0.5.0:
   dependencies:
     "@types/react" "^16.8.8"
 
-react-native-sensitive-info@~5.4.0:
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -13053,7 +13053,7 @@ react-native-safe-area@^0.5.0:
   dependencies:
     "@types/react" "^16.8.8"
 
-react-native-sensitive-info@~5.4.0:
+react-native-sensitive-info@~5.4.2:
   version "5.4.2"
```
Cauldrath: It sems like this PR won't affect anything, because yarn.lock was loading 5.4.2 already anyway.

bweissbart:
Closing per above comment.

PR #909, opened 2019-11-01
Merge develop into feature/rn-60

Cauldrath:
Includes https://github.com/brandingbrand/flagship/pull/907

PR #907, opened 2019-10-29
fix(fsapp): Basic React Native Navigation Web Wrapper

Cauldrath:
Not showing modals as modals (pushing the screen instead) and not running the slide menu, but most screen transitions should work.

I'm making this PR with this partial version so that anyone trying to run this branch can build navigation using the wrapper, which should be fully-functional for apps.

PR #915, opened 2019-11-07
Update to react-native 0.61.5 and Typescript 3.7.4

Cauldrath:
Includes:
https://github.com/brandingbrand/flagship/pull/856
https://github.com/brandingbrand/flagship/pull/907
https://github.com/brandingbrand/flagship/pull/909


Cauldrath:
Now includes https://github.com/brandingbrand/flagship/pull/831

Cauldrath:
Autolinking should now be happening.

Cauldrath:
Issue introduced by upgrading to RN61:
https://github.com/wix/react-native-navigation/issues/2331

This causes the app to crash instead of restart when switching to a different datasource through the dev menu on Android.

Cauldrath:
Updated most modules to the latest version.
Modules not updated to the latest:
react-native-permissions@2 changed how permissions are accessed, with separate handling for Android and iOS
path-to-regexp@4/5 changed the structure of where various functions are. This can be updated for iOS and Android, but on web react-router uses an old version and breaks by loading the new one.
@babel/runtime@7.6 doesn't work with tcomb-form-native.

Cauldrath:
Pulled in the latest code from feature/rn-60 and develop, added some backwards-compatibility functionality for react-native-navigation 2/3, and some error messaging for problems that could occur for users upgrading from Flagship 7.

Cauldrath:
Merged latest from feature/react-native-navigation-2

Cauldrath:
Upgrading to react-native 61:
https://react-native-community.github.io/upgrade-helper/?from=0.59.1&to=0.61.5
Note that a number of components were moved into their own repos.
All xib files will need to be converted to storyboard files.
All Android code should be converted to Android X or work with Jetifier.
If a node module does not work with the autolinker, it needs to be added to package.json in the array under flagship->forceLink. react-native-restart seems to be the only one in the basic project that needs it.

Upgrading to react-native-navigation 3:
https://wix.github.io/react-native-navigation/#/docs/top-level-api-migration

The main difference is that when you push or do similar screen transitions, it should look like:
```
navigator.push({
  component: {
    name: 'Home'
  }
});
```
when before it was
```
navigator.push({
  screen: 'Home'
});
```
Navigator should be imported from `@brandingbrand/fsapp` instead of `react-native-navigation`. A number of functions have been converted to work with react-native-navigation 3, but will give deprecation warnings instructing you how to use the updated syntax.
handleDeepLink and setOnNavigator event no longer are part of react-native-navigation, so you'll need to handle those transitions some other way and use the new events:
https://wix.github.io/react-native-navigation/#/docs/events

Cauldrath:
Pulled the latest from develop and upgraded react-native-navigation to 4.8.4.

PR #925, opened 2019-11-14
feat(flagship): Add option to only include the default env

Cauldrath:
Usage: `yarn run init --env prod --onlyDefault`

Will only include the default env, which should reduce the size of production bundles by excluding the environments that aren't needed.


Cauldrath:
Fixed JSDoc error.

PR #946, opened 2019-12-03
chore: update Podfile source to CocoaPods CDN

vancehu:
Background story: https://blog.cocoapods.org/CocoaPods-1.7.2/
Switching to this CDN saves up to 10 minutes per travis build across all of our projects


bweissbart:
This was re-created by #958 which makes the change in the feature/rn-60 branch. This is a breaking change (it requires updating to a new version of CocoaPods) so it shouldn't go directly into develop.

PR #950, opened 2019-12-13
feat(flagship): add Android App Bundle (.aab) Support

vancehu:
https://developer.android.com/guide/app-bundle
Added a new Fastfile lane: appcenter-app-bundle


packages/flagship/android/fastlane/Fastfile
```diff
@@ -23,6 +23,19 @@ lane :appcenter do
...
+# Use this lane if you want to generate app bundle: https://developer.android.com/guide/app-bundle
+lane :appcenter_app_bundle do
+  # build the release variant
+  gradle(task: "assembleBundle")
+
+  appcenter_upload(
+    #PROJECT_MODIFY_FLAG_appcenter_api_token
+    owner_name: "INJECTED_FROM_CONFIG", #PROJECT_MODIFY_FLAG_appcenter_owner_name
+    app_name: "INJECTED_FROM_CONFIG", #PROJECT_MODIFY_FLAG_appcenter_app_name_android
+    destination_type: "store"
```
NickBurkhartBB: @vancehu is this part for destination_type for store needed? It seems to be missing in our other fastlane paths so I wonder if we want to just use the default type group instead. Or is this a separate thing we would also need to configure?

vancehu: @NickBurkhartBB Yes. appcenter will throw an error if app bundle destination_type is not set to "store"

vancehu: https://github.com/microsoft/fastlane-plugin-appcenter/blob/master/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb

NickBurkhartBB: ok, thanks, I was having trouble finding specific docs, but code works

PR #954, opened 2019-12-16
feat(fsengagement): add animated card list for demo

andyschwob:
This adds the following component to fsengagement for rendering of the inbox:

ImageCarousel
Divider
RoundedImageCard - specific card that will be animated in the card feed
InboxWrapper - wraps components in animated stories with a white background (needed as they are displayed in a modal over title card)
DemoProductCarousel - Product Carousel that opens up a scrollable modal carousel when tapped (for demo purposes)

Some of the other base files and styles were imported from the RNN2 feature branch.


bweissbart:
Travis is timing out installing pods. I can't get it to fix this, nor can I reproduce locally. As such I'm going to merge.

PR #980, opened 2020-01-15
feat(fscomponents): bb cms feedback form

sentiurin:
### Description:
- We already have a version of that in flagship. See here: https://github.com/brandingbrand/flagship/blob/develop/packages/fscomponents/src/components/CMSFeedback.tsx
- We want to create another "*CMSFeedbackForm*" that is similar to the screenshot below. 
- Requirements: 
* Need to create a form that consists of a title, a textbox and a CTA
* The title fontSize, fontColor and font needs to be customizable
* The textbox border, height, and font need to be customizable as well
* The CTA title, background color, border, font need to be customizable as well
* Please keep same functionality as the existing form above 
* Please don't remove the old form - This would be an additional option

### Screenshot:
Screenshot 2020-01-13 at 17 00 31

### P.S.
I was Integrate "Formik" library into "fscomponents".
CMS Feedback Form created based on "Formik" library.
CMS Feedback Form retained its functionality as in the CMS Feedback component.
CTA title, textbox and title may have custom styles.

### DIFF
--- a/packages/fscomponents/package.json
+++ b/packages/fscomponents/package.json
@@ -19,6 +19,7 @@
     "@react-native-community/async-storage": "^1.6.1",
     "credit-card": "^3.0.1",
     "decimal.js": "^10.0.1",
+    "formik": "^2.1.1",
     "lodash-es": "^4.17.10",
     "memoize-one": "^4.0.2",
     "photoswipe": "^4.1.2",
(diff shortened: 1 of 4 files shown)


Cauldrath:
You have unresolved conflicts in yarn.lock

bweissbart:
Closing as this hasn't been needed.

PR #996, opened 2020-01-17
refactor(fscomponents): FLAGSHIP-62 update to function component

huizhuz:
### Ticket
https://jira.brandingbrand.com/browse/FLAGSHIP-62

### Description
SelectableList used to be a class-based component. This PR is to update it to a function component.


packages/fscomponents/src/components/SelectableList.tsx
```diff
@@ -22,28 +22,26 @@ const S = StyleSheet.create({
   }
 });
 
-export class SelectableList extends Component<SelectableListProps> {
-  renderItem = (item: SelectableListItem, index: number) => {
+export const SelectableList = (props: SelectableListProps): JSX.Element => {
```
bweissbart: You can type this as a Function Component as follows:

```
export const SelectableList: React.FunctionComponent<SelectableListProps> = props => { ... }
```

PR #1018, opened 2020-01-22
chore(fstestproject): get test project working again

bweissbart:
Added up-to-date metro and babel configs and missing dependencies so the test project works. To run: yarn testproject:init && yarn testproject:run-ios

### DIFF
--- a/packages/fstestproject/metro.config.js
+++ b/packages/fstestproject/metro.config.js
@@ -14,7 +14,7 @@ function generateBlacklist() {
 
   fs.readdirSync(packageModules).forEach(moduleDir => {
     if (moduleDir === 'react-native') {
-      // match all packages except pirateship
+      // match all packages except fstestproject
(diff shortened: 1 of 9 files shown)


packages/fstestproject/metro.config.js
```diff
@@ -13,11 +14,11 @@ function generateBlacklist() {
 
   fs.readdirSync(packageModules).forEach(moduleDir => {
     if (moduleDir === 'react-native') {
-      // match all packages except fstestproject
-      duplicateModules.push(new RegExp('packages/(?:(?!fstestproject).*)/node_modules/' + moduleDir + '/.*'));
-      duplicateModules.push(new RegExp(rootModules + '/' + moduleDir + '/.*'));
+      // match all packages except pirateship
```
Cauldrath: Should be "fstestproject" still.

bweissbart: Good catch. Fixed.

PR #1023, opened 2020-01-23
feat(fscomponents): FLAGSHIP-102 accessibility for Filter

richter31:


packages/fscomponents/src/components/FilterList/FilterListDrilldown.tsx
```diff
@@ -347,6 +351,9 @@ export class FilterListDrilldown extends PureComponent<
         <TouchableOpacity
           style={S.secondLevelHeader}
           onPress={this.backToFirstLevel}
+          accessibilityRole={'button'}
+          accessibilityHint={'Go back one filter level'}
```
Cauldrath: It seems like this and "Toggles Filter" above should be in the translation file, too.

richter31:
@Cauldrath  updated

PR #1036, opened 2020-01-29
fix(fscomponents): Name inner component functions

Cauldrath:
Because these functions weren't named, they show up as "Unknown" in the react-native debugger. This gives those functions a name so you can identify them easier.


packages/fscomponents/src/components/CategoryBox.tsx
```diff
@@ -55,7 +55,9 @@ export const CategoryBox = React.memo((props: CategoryBoxProps): JSX.Element =>
       </View>
     </TouchableHighlight>
   );
-}) as any;
+};
+
+export const CategoryBox = React.memo(CategoryBoxInner) as any;
```
bweissbart: Do we need the `as any` type coercion here?

PR #1061, opened 2020-02-11
feat(fsbazaarvoice): Expand review data source

varzaman:
Added new optional params to the bazaarvoice review data source while leaving existing functionality in place.
Filter allows for multiple values following the "Filter=first&Filter=second&Filter=third" format.

Found no existing framework to create a test case in flagship. In the interest of time, I tested it directly within the CVS app.

### DIFF
--- a/packages/fsbazaarvoice/src/BazaarvoiceDataSource.ts
+++ b/packages/fsbazaarvoice/src/BazaarvoiceDataSource.ts
@@ -7,6 +7,7 @@ import {
 import * as BazaarvoiceNormalizer from './BazaarvoiceNormalizer';
 import * as BazaarvoiceDenormalizer from './BazaarvoiceDenormalizer';
 import { BazaarvoiceReviewRequest } from './BazaarvoiceReviewRequest';
+import qs from 'qs';
 
 export interface BazaarvoiceConfig {
   endpoint: string;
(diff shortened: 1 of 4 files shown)


varzaman:
Looks like the test for FSNetwork is failing and it doesn't seem related to anything I touched.

packages/fsbazaarvoice/src/BazaarvoiceDataSource.ts
```diff
@@ -31,18 +32,26 @@ export class BazaarvoiceDataSource extends AbstractReviewDataSource implements R
...
+    const filter = query.filter || `ProductId:${id}`;
+
     const params: BazaarvoiceReviewRequest = {
-      Filter: `ProductId:${id}`,
+      Filter: filter,
       Include: 'Products',
       Stats: 'Reviews',
-      Limit: query.limit || 10
+      Limit: query.limit || 10,
+      ...(query.sort ? {Sort: query.sort} : {})
```
Cauldrath: Why not just make it `Sort: query.sort`? Is it shoving in "Sort=undefined" that way?

varzaman: I wasn't sure if it would still add an undefined parameter or not.

I just checked an qs indeed omits them, so I can change this.

PR #1064, opened 2020-02-14
refactor(fscommerce): Make LocationItem a function component

CBoetto:
[Ticket](https://jira.brandingbrand.com/browse/FLAGSHIP-54)

**Description:**
We should audit all of the "vanilla" React components within Flagship (i.e., implemented as class components that extend from React.Component) to determine whether they can be converted to pure components or function components.

There's a writeup in Tettra explaining when the different component types should be used: https://app.tettra.co/teams/brandingbrand/pages/react-component-optimization-strategies

Please read the above article, but to summarize:

Function components are the most performant and should be used for simple components that don't need access to the full component API. Hooks such as useState and useEffect allow for some additional functionality. If the component is pure, and is expected to be called with the same props, wrap it in React.memo()
Pure components have access to the full component feature set but won't get re-rendered unless their props or state change based on a shallow comparison. These should be used if the component and its children exhibit pure behavior.
If a component or its children doesn't exhibit pure behavior, it should remain a vanilla React.Component.


bweissbart:
@CBoetto

CBoetto:
@bweissbart Sorry for the delay on this.

Also, I was running into some trouble with storybook but when I could get storybook to work for me, LocationItem looked like it was functioning as intended now.

PR #1087, opened 2020-03-18
feat(fscomponents): Add styling and prop options to ZoomCarousel

varzaman:
-Add prop to hide page indicators
-Fix issue where if fillContainer is set to true and showThumbnails is set to true, thumbnails won't show
-Add prop for custom styling for the fillContainer prop
-Add two stories to the storybook showing fillContainer prop for ZoomCarousel

### DIFF
--- a/packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.web.tsx
+++ b/packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.web.tsx
@@ -306,8 +306,7 @@ export class ZoomCarousel extends Component<ZoomCarouselProps, ZoomCarouselState
 
   render(): JSX.Element {
     const { peekSize = 0, gapSize = 0 } = this.props;
-    const height =
-      this.props.fillContainer ?
+    const height = this.props.fillContainer ? this.props.fillContainerStyle ||
       (this.props.showThumbnails ? S.fullHeightWithThumbnails : S.fullHeight)
        : null;
 
@@ -322,7 +321,7 @@ export class ZoomCarousel extends Component<ZoomCarouselProps, ZoomCarouselState
         >
           <div
             id={`zoom-carousel-${this.id}`}
-            style={this.props.fillContainer ? S.fullHeight : undefined}
+            style={this.props.fillContainer ? {height: '100%'} : undefined}
           >
             <MultiCarousel
               ref={this.extractMultiCarousel}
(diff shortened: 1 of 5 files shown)


packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.web.tsx
```diff
@@ -303,18 +306,23 @@ export class ZoomCarousel extends Component<ZoomCarouselProps, ZoomCarouselState
...
         onLayout={this.handleLayoutChange}
       >
         <View
-          style={this.props.fillContainer ? S.fullHeight : null}
+          style={height}
         >
           <div
             id={`zoom-carousel-${this.id}`}
-            style={this.props.fillContainer ? {height: '100%'} : undefined}
+            style={this.props.fillContainer ? S.fullHeight : undefined}
```
bweissbart: I tried running StoryBook and got a typing error on your changes here. I think the problem is that RN web doesn't know how to handle a RN StyleSheet element inside of a HTML element. You can either revert this specific change to fix the error, or you can change `<div>` to the following:

```
<View
  nativeID={`zoom-carousel-${this.id}`}
  style={this.props.fillContainer ? S.fullHeight : undefined}
>
```

Note that `nativeID` will become `id` when rendered as HTML.

packages/fscomponents/src/components/ZoomCarousel/types.ts
```diff
@@ -44,4 +44,18 @@ export interface ZoomCarouselProps {
...
+   * @example true
+   */
+  hidePageIndicator?: boolean;
+
+  /**
+   * The styling used if fill container is set to true (that makes all carousel elememts expand)
+   *
+   * @example {{height: '50%'}}
+   */
+  fillContainerStyle?: StyleProp<ViewStyle>;
```
bweissbart: This doesn't appear to be utilized anywhere in the ZoomCarousel code.

varzaman: Resolved here: https://github.com/brandingbrand/flagship/pull/1087/files#diff-badb4df4ea5cc3f25061a62fb1d31be9R309

packages/fscomponents/src/components/ZoomCarousel/ZoomCarousel.web.tsx
varzaman: Reverted back

PR #1093, opened 2020-03-20
feat(fscomponents): Add image counter, new style options to ZoomCarousel

varzaman:
- Added a new option that can be rendered: image counter
 --Absolutely placed component with custom rendering or custom styles as props
 - Added new prop options for the carousel and image containers
 - changed the default container styling to be flex friendly, fixing an issue where thumbnails would overflow beyond parent boundaries

Added some storybook options and also tested as a linked package inside of a test app.

### DIFF
--- a/packages/fscomponents/src/components/ZoomCarousel/types.ts
+++ b/packages/fscomponents/src/components/ZoomCarousel/types.ts
@@ -53,9 +53,39 @@ export interface ZoomCarouselProps {
   hidePageIndicator?: boolean;
 
   /**
-   * The styling used if fill container is set to true (that makes all carousel elememts expand)
+   * The styling of the container that holds the image carousel and thumbnails
    *
-   * @example {{height: '50%'}}
+   * @example {{flex: 1}}
    */
-  fillContainerStyle?: StyleProp<ViewStyle>;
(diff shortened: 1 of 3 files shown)


packages/fscomponents/src/components/ZoomCarousel/types.ts
```diff
@@ -53,9 +53,39 @@ export interface ZoomCarouselProps {
...
 
   /**
-   * The styling used if fill container is set to true (that makes all carousel elememts expand)
+   * The styling of the container that holds the image carousel and thumbnails
    *
-   * @example {{height: '50%'}}
+   * @example {{flex: 1}}
    */
-  fillContainerStyle?: StyleProp<ViewStyle>;
+  carouselContainerStyle?: StyleProp<ViewStyle>;
```
bweissbart: It doesn't look like this is used anywhere.

varzaman: https://github.com/brandingbrand/flagship/pull/1093/files#diff-badb4df4ea5cc3f25061a62fb1d31be9R362 

line 362

varzaman: disregard I cannot read

PR #1095, opened 2020-03-23
feat(fsengagement): add more event emitters and position data

andyschwob:
- Adds the event emitter `swipeCard` when using a FullScreenImageCarousel card, emits an event when swiping to each new card.
- If available, send the position of the card with the events `viewLink` and `viewStory`


deemaabdallah:
@calebabrahamBB See this change

PR #1102, opened 2020-03-26
feat(fscomponents): ShareButton web support

Cauldrath:
react-native-web has support for the same Share api that react-native uses, but it isn't supported by all browsers (mostly just mobile ones):
https://caniuse.com/#feat=web-share

This will attempt to use that same api and, if it isn't available, attempt to copy the url to the clipboard.

### DIFF
--- a/packages/fscomponents/src/components/ShareButton/ShareButton.tsx
+++ b/packages/fscomponents/src/components/ShareButton/ShareButton.tsx
@@ -2,8 +2,8 @@ import React, { FunctionComponent, memo } from 'react';
 import { Clipboard, Image, Share, TouchableOpacity } from 'react-native';
 import { Alert } from '../Alert';
 import { ShareButtonProps } from './ShareButtonProps';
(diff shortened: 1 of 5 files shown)


packages/fscomponents/src/components/ShareButton/ShareButton.tsx
```diff
@@ -1,22 +1,50 @@
...
+          // @ts-ignore ShareContent has a version with url and one without
+          const { url } = props.content;
+          // @ts-ignore Web returns false on failure, native returns void
+          if (Clipboard.setString(url) !== false) {
+            Alert.alert('URL copied to clipboard: ' + url);
+          } else {
+            Alert.alert('Unable to copy the URL to the clipboard');
+          }
+        } else {
+          Alert.alert('Message sharing is not supported by your browser.');
```
bweissbart: Can you please add translations for these three alert dialogs?

PR #1139, opened 2020-04-14
feat: add story to ChangePassword component

Bodrus:
ChangePasswordStory has been added.


.idea/workspace.xml
```diff
@@ -0,0 +1,502 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bweissbart, line 1: Can you please remove this file from your commit?

PR #1144, opened 2020-04-15
feat: add story to ChangePassword component

Bodrus:
feat: add story to ChangePassword component


bweissbart (changes requested):
If I run the story, click the Current Password field, then click another field, I get a "cannot read property 'length' of null" error.

Cauldrath (approved):
Seems good.

PR #1194, opened 2020-04-21
feat: Add serializable props for AddToCart component

IvanMatchin:
This gives us a serializable version of the AddToCart properties. This allows us to generate a schema of the properties that can be represented in JSON.


bweissbart:

PR #1195, opened 2020-04-21
feat: add serializable props for FadeInImage component

romaniukua:
Create serializable props interface for FadeInImage component


bweissbart:

PR #1202, opened 2020-04-22
feat: add serializable props for MultiCarousel component

romaniukua:
Create serializable props interface for MultiCarousel component


bweissbart:

PR #1204, opened 2020-04-22
feat: add serializable props to ProductItem component

Bodrus:
Create serializable props interface for ProductItem component.


bweissbart:

PR #1226, opened 2020-04-23
feat: Add serializable props for EmailForm component

IvanMatchin:
This gives us a serializable version of the EmailForm properties. This allows us to generate a schema of the properties that can be represented in JSON.


packages/fscomponents/src/components/EmailForm.tsx
```diff
@@ -12,14 +12,17 @@ export interface EmailFormValue {
   email: string;
 }
 
-export interface EmailFormProps {
+export interface SerializedEmailFormProps {
+  value?: EmailFormValue;
+}
+
+export interface EmailFormProps extends Omit<SerializedEmailFormProps, any> {
```
bweissbart, line 19: Since we're using all of the props we don't need Omit here.

bweissbart:

PR #1236, opened 2020-04-23
feat: add serializable props for ReviewItem

romaniukua:
Create serializable props interface for ReviewItem component


bweissbart:

PR #1243, opened 2020-04-23
feat(fsi18n): Replace all occurrences of "any" in FSI18N package

olkorsia:
Replace all occurrences of "any" in FSI18N package


packages/fsi18n/src/i18nHelper.ts
```diff
@@ -240,10 +240,10 @@ export default class I18nHelper {
    * @returns {boolean} - Whether or not the value is a translation key
    */
   protected isTranslationKey(
-    translationKey: any
```
bweissbart, line 243: I think we can leave this as `any` because I believe the intention of this method is to take any value and return whether it's a key or not.

bweissbart (changes requested):

alexandrvlasov:
@bweissbart I changed.

PR #1244, opened 2020-04-23
feat: add serializable props for swatches

rosahbruno:
## Description
Add serializable props for Swatches component


Cauldrath:
There should be no required properties that aren't serializable. `items` should have been part of the serializable props.

Cauldrath:
Fixed here: https://github.com/brandingbrand/flagship/pull/1497

PR #1253, opened 2020-04-24
feat: add serializable props to ApplePayButton component

Bodrus:
add serializable props to ApplePayButton component


bweissbart:

PR #1287, opened 2020-04-29
feat(flagship): iOS Multiple Target Support

alexrisch:
Added Xcode package, tried to keep usage to a minimum
Updated types to include IOSExtensions
Added iosExtensions function
Function will do some basic setup for targets that is needed when submitting
Updates Fastlane with extension profile
Updates bundle versions to match app
Use of Xcode lib:
Add target
update project build phases
Add frameworks to target
Adds files in extension to project group (makes them visible in Xcode)
Saves project file

Then custom scripts will add necessary build settings this includes
entitlements
provisioning profiles
code sign identities

Finally it will replace other set strings after copying target into project

Testing:
Take an existing project that uses extension and update the env as needed
reinit the project and test the extension


bweissbart:
@deemaabdallah - can you please review the general flow, especially the code signing bits?

Cauldrath (approved):
Code looks good to me (maybe it could be optimized some to not use as many sync functions?) and I am still able to run PirateShip in iOS.

Cauldrath (approved):
Seems good.

PR #1298, opened 2020-04-30
chore(fscomponents): Remove tcomb-form-native

Cauldrath:
This will allow us to update to newer versions of babel. @babel/core 7.7.0 and babel-loader 8.1.0 caused the fsshopify tests to fail, so we're still using the version before those. Some of the functionality in PirateShip had to be commented-out until a Formik version can be implemented, but those pages don't appear to be getting used anywhere.


Cauldrath:
This was merged and reverted in other PRs.

PR #1318, opened 2020-05-05
feat(fscomponents): Replace all occurrences of "any" in Carousel

olkorsia:
## Description
Replace all occurrences of "any" in Carousel


packages/fscomponents/src/components/Carousel/CarouselProps.ts
```diff
@@ -1,12 +1,13 @@
...
 import { StyleProp, ViewStyle } from 'react-native';
+import Swiper from 'react-native-swiper';
 
 // TODO: Unify these web/native options and clean up these types ...a lot
 export interface CarouselProps {
   currentPageIndicatorColor?: string;
   height?: number;
   loop?: boolean;
-  nativeOptions?: any;
+  nativeOptions?: Swiper;
```
bweissbart, line 10: I think this needs to be `SwiperProperties` from `@types/react-native-swiper`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native-swiper/index.d.ts#L32

Using `Swiper` will expect an instance of the component instead.

bweissbart (changes requested):

alexandrvlasov:
@bweissbart I can't update these props to @types/react-native-swiper library, because react-native-swiper library contains index.d.ts file with types. And If we will be using the @types library its can have different properties.

PR #1321, opened 2020-05-05
feat(fscomponents): replace all occurrences of "any" in FilterList

romaniukua:


packages/fscomponents/src/components/Accordion.tsx
```diff
@@ -141,7 +141,7 @@ export interface AccordionProps extends Omit<
    * Content of the accordion
    * @deprecated Make the contents a child instead
    */
-  content?: JSX.Element;
+  content?: JSX.Element | JSX.Element[];
```
bweissbart, line 144: You can change this to `React.ReactNode` to cover all use cases.

bweissbart (approved):

PR #1338, opened 2020-05-06
feat(fscomponents): Replace all occurrences of "any" in SingleLineForm component

MatchinIvan:
### Description
Replace all occurrences of "any" in SingleLineForm component


packages/fscomponents/src/components/SingleLineForm.tsx
```diff
@@ -1,19 +1,19 @@
...
-  onSubmit?: (value: any) => void;
-  submitButtonStyle?: any;
-  submitTextStyle?: any;
-  submitText?: any;
-  value?: any;
-  style?: any;
+  onSubmit?: <T>(value: T) => void;
+  submitButtonStyle?: StyleProp<ViewStyle>;
+  submitTextStyle?: StyleProp<ViewStyle>;
+  submitText?: () => void;
```
bweissbart, line 14: These should be `StyleProp<TextStyle>` and `Text`.

packages/fscomponents/src/components/SingleLineForm.tsx
```diff
@@ -1,19 +1,19 @@
...
+  value?: string;
```
bweissbart, line 15: We should check but I think this is going to be an object since the list of fields is an object.

bweissbart (changes requested):

PR #1339, opened 2020-05-06
feat(fscomponents): Replace all occurrences of "any" in EmailForm component

MatchinIvan:
### Description 

Replace all occurrences of "any" in EmailForm component


packages/fscomponents/src/components/EmailForm.tsx
```diff
@@ -6,31 +6,33 @@ import * as t from 'tcomb-form-native';
...
-  fieldsStyleConfig?: any;
+  fieldsStyleConfig?: Dictionary;
   labelPosition?: FormLabelPosition;
   onSubmit?: (value: EmailFormValue) => void;
-  submitButtonStyle?: any;
-  submitTextStyle?: any;
-  submitText?: any;
+  submitButtonStyle?: StyleProp<ViewStyle>;
+  submitTextStyle?: StyleProp<ViewStyle>;
+  submitText?: StyleProp<ViewStyle>;
```
bweissbart, line 23: `submitTextStyle` should be `StyleProp<TextStyle>` and `submitText` should be `Text`. These types come from here: https://github.com/brandingbrand/flagship/blob/develop/packages/fscomponents/src/components/SingleLineForm.tsx ; ideally we should clean up the types on that component and then reference them here.

bweissbart (changes requested):

PR #1374, opened 2020-05-08
feat(any-occurrences): Replace all occurrences of "any…

olexandor:
### Description:

Replace all occurrences of "any" in DigitalWalletProvider

### DIFF
--- a/packages/fscomponents/src/components/DigitalWalletProvider.tsx
+++ b/packages/fscomponents/src/components/DigitalWalletProvider.tsx
@@ -44,7 +44,7 @@ export function withDigitalWallet<P extends {}>(
 
   return class DigitalWalletProvider extends React.Component<ResultProps, WithDigitalWalletState> {
     applePay: ApplePay;
-    promises: Promise<any>[] = [];
+    promises: Promise<void>[] = [];
 
     constructor(props: ResultProps) {
       super(props);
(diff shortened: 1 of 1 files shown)


packages/fscomponents/src/components/DigitalWalletProvider.tsx
```diff
@@ -44,7 +44,7 @@ export function withDigitalWallet<P extends {}>(
 
   return class DigitalWalletProvider extends React.Component<ResultProps, WithDigitalWalletState> {
     applePay: ApplePay;
-    promises: Promise<any>[] = [];
+    promises: Promise<boolean>[] = [];
```
bweissbart, line 47: `initApplePay` doesn't return anything so this should be `Promise<void>`.

packages/fscomponents/src/components/DigitalWalletProvider.tsx
```diff
@@ -77,7 +77,7 @@ export function withDigitalWallet<P extends {}>(
         });
     }
 
-    async initApplePay(): Promise<any> {
+    async initApplePay(): Promise<boolean> {
```
bweissbart, line 80: `Promise<void>`.

bweissbart (changes requested):

packages/fscomponents/src/components/DigitalWalletProvider.tsx
olexandor: @bweissbart Changes request implemented.

packages/fscomponents/src/components/DigitalWalletProvider.tsx
olexandor: Changes request implemented.

PR #1397, opened 2020-05-12
feat(pirateship): Replace all occurrences of "any" in PSHalfModal

MatchinIvan:
### Description
Replace all occurrences of "any" in PSHalfModal


packages/pirateship/src/components/PSHalfModal.tsx
```diff
@@ -13,7 +13,7 @@ export interface PSHalfModalProps {
   visible: boolean;
   renderLeftItem?: () => JSX.Element;
   renderRightItem?: () => JSX.Element;
-  children?: any;
+  children?: JSX.Element;
```
bweissbart, line 16: You should be able to remove this property altogether -- `children` should come out-of-the-box with the `FunctionComponent` interface.

bweissbart (changes requested):

PR #1411, opened 2020-05-14
chore(flagship): Replace all occurrences of any in Flagship package

sentiurin:


packages/flagship/src/lib/rename.ts
```diff
@@ -99,7 +99,7 @@ export function files(oldName: string, newName: string, ...pathComponents: strin
     // Rename each path
     results.forEach(oldPath => {
       // Only replace the final occurence in the path
-      let newPath: any = oldPath.split('/');
+      let newPath: string[] | string = oldPath.split('/');
```
bweissbart, line 102: The return type for string.split is always going to be `string[]`.

bweissbart (dismissed):

PR #1431, opened 2020-05-19
feat(any-occurrences): Replace all occurrences of "any" in packages/f…

olexandor:
### Description:

Replace all occurrences of "any" in packages/fscomponents/src/lib/wallets

### DIFF
--- a/packages/fscomponents/src/lib/wallets/ApplePay.web.ts
+++ b/packages/fscomponents/src/lib/wallets/ApplePay.web.ts
@@ -1,29 +1,42 @@
 import { default as ApplePayBase } from './ApplePay.base';
 
+export interface ApplePaySession {
+  openPaymentSetup?: (identifier: string) => void;
+  canMakePaymentsWithActiveCard?: (identifier: string) => void;
+}
+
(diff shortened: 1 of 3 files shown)


packages/fscomponents/src/lib/wallets/ApplePay.web.ts
```diff
@@ -1,29 +1,42 @@
 import { default as ApplePayBase } from './ApplePay.base';
 
+export interface ApplePaySession {
+  openPaymentSetup?: (identifier: string) => void;
+  canMakePaymentsWithActiveCard?: (identifier: string) => void;
```
bweissbart, line 5: Per Apple's docs, both of these methods return Promise<boolean>.

https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession

olexandor: @bweissbart Type for returned value from methods 'openPaymentSetup' and 'canMakePaymentsWithActiveCard' was changed. Please check.

bweissbart (approved):

PR #1475, opened 2020-06-08
fix(pirateship): svg arrows no longer black boxes

erichfeinstein:
- Removing the return statement in the render image function seems to remove the black boxes and render the proper arrows

### DIFF
--- a/packages/pirateship/src/components/PSRow.tsx
+++ b/packages/pirateship/src/components/PSRow.tsx
@@ -48,7 +48,7 @@ export interface PSRowProps {
 const PSRow: FunctionComponent<PSRowProps> = (props): JSX.Element => {
   const renderImage = (): JSX.Element => {
     if (props.renderImage) {
-      return props.renderImage();
+      props.renderImage();
     }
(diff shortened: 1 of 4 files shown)


packages/pirateship/src/components/PSRow.tsx
```diff
@@ -48,7 +48,7 @@ export interface PSRowProps {
 const PSRow: FunctionComponent<PSRowProps> = (props): JSX.Element => {
   const renderImage = (): JSX.Element => {
     if (props.renderImage) {
-      return props.renderImage();
```
bweissbart, line 51: This should return because it's intended to short-circuit the default Image render with a custom render function. If something in PirateShip is providing a renderImage prop using SVGs, we should just remove that custom render function so we defer to the default behavior.

erichfeinstein: I see. Maybe I should take some more time to dive into this function. Removing that if statement from 50-52 got it working on the problematic device and iOS version

PR #1502, opened 2020-06-29
fix(pirateship): pressing X removes items from cart

erichfeinstein:
- Resolved issue where pressing the delete item button on an item in your cart would not remove it
- Fixed issue where cart count navbar badge would not properly update when removing items


Cauldrath (approved):
Seems good.

PR #1506, opened 2020-07-02
fix: engagement web-url button crashes fix

vancehu:
two types of crashes when user tries to click on a `learn more` button:
image

- unrecognized selector sent to instance 0x.....
    - payload passed to RN bridge is not using the correct structure. Fixed: `{ color: '#f5f2ee' }`

image
- cannot read property `bindNavigation` of undefined:
   - navigator object is missing. Fixed by passing navigator to passProps


bweissbart:
@vancehu - tsc error

```
src/EngagementComp.tsx(26,22): error TS6133: 'OptionsTopBarBackground' is declared but its value is never read.
```

vancehu:
updated

bweissbart:
This was fixed in another PR done by @Cauldrath.

PR #1534, opened 2020-07-20
fix(fscomponents): add additional safety check for ProductItemImage

nathan-sankbeil:
Component returns null if an image source lacks a uri


Cauldrath:
This seems like a case where the image shouldn't be defined if it isn't a valid source, so it's fixing a symptom to the problem instead of the problem itself, and it also makes it so we only support the source format where it uses uri.

PR #1535, opened 2020-07-20
fix(fscomponents): Fix crashes with bad currency codes

Cauldrath:
If your currency code was invalid, it would crash the app when it tried to render. This wraps those conversions in a try/catch, so it will put the error in the console and display as if the currency value was undefined instead of crashing.

### DIFF
--- a/packages/fscomponents/src/components/CartItem/CartItemDetails.tsx
+++ b/packages/fscomponents/src/components/CartItem/CartItemDetails.tsx
@@ -36,20 +36,38 @@ memo((props): JSX.Element => {
     totalPrice
   } = props;
 
+  let convertedPrice: string | undefined;
+  try {
+    if (price) {
+      convertedPrice = FSI18n.currency(price);
+    }
+  } catch (e) {
+    console.error(e);
+  }
+
+  let convertedTotalPrice: string | undefined;
+  try {
+    if (totalPrice) {
+      convertedTotalPrice = FSI18n.currency(totalPrice);
+    }
+  } catch (e) {
+    console.error(e);
+  }
+
   return (
     <View style={style}>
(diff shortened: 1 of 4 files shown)


packages/fscomponents/src/components/CartItem/CartItemDetails.tsx
```diff
@@ -36,20 +36,38 @@ memo((props): JSX.Element => {
...
   return (
     <View style={style}>
       {itemText && <Text style={[defaultStyles.detailText, detailTextStyle]}>{itemText}</Text>}
       {productId && <Text style={[defaultStyles.detailText, detailTextStyle]}>#{productId}</Text>}
-      {price && (
+      {convertedPrice && (
         <Text style={[defaultStyles.detailText, detailTextStyle]}>
           {FSI18n.string(componentTranslationKeys.unitPrice)}:
-          {FSI18n.currency(price)}
+          {FSI18n.currency(convertedPrice)}
```
nathan-sankbeil, line 64: I think you forgot to update this.

packages/fscomponents/src/components/CartItem/CartItemDetails.tsx
```diff
@@ -36,20 +36,38 @@ memo((props): JSX.Element => {
...
         </Text>
       )}
-      {totalPrice && (
+      {convertedTotalPrice && (
         <Text style={detailTextStyle}>
           {FSI18n.string(componentTranslationKeys.totalPrice)}:
-          {FSI18n.currency(totalPrice)}
+          {FSI18n.currency(convertedTotalPrice)}
```
nathan-sankbeil, line 70: I think you forgot to update this.

PR #1543, opened 2020-07-24
fix(flagship): iOS Pods not being added

alexrisch:
Our logic to add pods was broken because we removed the identifying line
Readded line for Podfile updates

Also updated logic for fs.update to use doesKeywordExist function to find other issues
This will allow us to better audit script logic in the future and find potential breaking changes sooner

### DIFF
--- a/packages/flagship/src/lib/fs.ts
+++ b/packages/flagship/src/lib/fs.ts
@@ -73,7 +73,7 @@ export function update(path: string, oldText: string | RegExp, newText: string):
 export function doesKeywordExist(path: string, keyword: string | RegExp): boolean {
   // TODO: This should use a streaming buffer
   const fileContent = readFileSync(path, { encoding: 'utf8' });
-  if (typeof(keyword) === 'string') {
+  if (typeof keyword === 'string') {
     return fileContent.indexOf(keyword) > -1;
   } else {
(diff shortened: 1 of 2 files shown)


deemaabdallah:
@bweissbart this is needed urgently! can you please take  a look?

packages/flagship/src/lib/fs.ts
```diff
@@ -65,8 +70,12 @@ export function update(path: string, oldText: string | RegExp, newText: string):
  * @param {string} keyword The keyword to search for in the file
  * @returns {boolean} True if the keyword exists in the file.
  */
-export function doesKeywordExist(path: string, keyword: string): boolean {
+export function doesKeywordExist(path: string, keyword: string | RegExp): boolean {
   // TODO: This should use a streaming buffer
   const fileContent = readFileSync(path, { encoding: 'utf8' });
-  return fileContent.indexOf(keyword) > -1;
+  if (typeof(keyword) === 'string') {
```
bweissbart, line 76: This should just be `if (typeof keyword === 'string') {`

JeromeFitz: Tagging a commit on @alexrisch behalf. Coming right up @bweissbart.

PR #1576, opened 2020-08-14
fix(fsengagement): fix page counter not fading away

andyschwob:
At some point in the transition to FS9 we ended up with two instances of the page counter view, one of which was not animated.  This made it appear that the page counter was not fading away when tapping into a story.


crherman7:
@deemaabdallah @andyschwob @bweissbart let's try to get this prioritized looks like we just need to get this reviewed / merged / released at this point! Thanks for finding this fix 🔥

PR #1594, opened 2020-08-18
feat: Update FS component to allow custom header to be passed in

Bodrus:
## Description
Update FS component to allow custom header to be passed in


bweissbart (changes requested):
@Bodrus - I believe what is actually needed here is for the ability to pass in an entirely different header component. I'd recommend adding a new optional prop to EngagementComp (called something like `renderHeader` that accepts a function that returns JSX. If this prop is provided, it will call that function instead of rendering the default "welcome" header.

Bodrus:
@bweissbart Changed and tested! Please look

packages/fsengagement/src/EngagementComp.tsx
```diff
@@ -455,11 +456,13 @@ export default function(
...
-      const name = this.props.headerName || '';
+      const { json, headerName } = this.props;
+      const name = headerName || '';
+      const headerTitleStyle = json && json.headerTitleStyle || {};
       const comma = name ? ', ' : '';
       return (
-        <Text style={styles.headerName}>
-          HELLO{comma}{name.toUpperCase()}
+        <Text style={[styles.headerName, headerTitleStyle]}>
+          Hello{comma}{name}
```
Cauldrath, line 465: While this does look good overall, this is changing the default behavior of the component. Adding `textTransform: uppercase` to styles.headerName should make it act the same.

Bodrus:
@Cauldrath Fixed!

Cauldrath (approved):
Seems good.

PR #1623, opened 2020-09-03
fix: resolve android build tools in subprojects

bweissbart:
This adds a block to build.gradle to set the Android build tools version for subprojects (from external dependencies). This addresses a build error in release mode in which libraries are attempting to reference Material UI. This appears to be the community-supported resolution to this problem: https://stackoverflow.com/a/53371395

This also runs Jetify whenever init runs because this is a required step for builds to work.

### DIFF
--- a/packages/flagship/android/build.gradle
+++ b/packages/flagship/android/build.gradle
@@ -22,6 +22,25 @@ buildscript {
     }
 }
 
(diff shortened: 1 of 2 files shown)


nathan-sankbeil (approved):
Seems good.

PR #1652, opened 2020-09-15
feat: FS-1314 - upgrade typescript to 4.0.2

bweissbart:
BREAKING CHANGE: This upgrades Typescript from 3.x to 4.x which includes changes that are for edge cases but still breaking nonetheless. The full changelog can be found here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#breaking-changes

Fixed a couple instances where TypeScript recognized that we were always overriding variables by using the spread operator. Also addressed a couple places where we were importing `default` even though no such export existed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "ts-jest": "^24.1.0",
     "ts-loader": "^6.0.0",
     "typedoc": "^0.15.0",
-    "typescript": "^3.7.2",
+    "typescript": "^4.0.2",
     "webpack": "^4.41.2"
   },
   "greenkeeper": {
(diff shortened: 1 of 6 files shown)


nathan-sankbeil (changes requested):
Seems good aside from the lodash thing.

bweissbart:
@nathan-sankbeil - with a clearer head I made updates that don't require lodash. I opted to keep the functionality as it existed before, with props taking precedence over parameters in the web screen wrapper.

nathan-sankbeil (approved):
Seems good.

PR #1656, opened 2020-09-16
feat: upgrade to RN 0.63

nathan-sankbeil:
upgrades to the latest version of react-native

BREAKING CHANGE: major react native update


bweissbart:
@nathan-sankbeil - Android stalled out on Travis, which for some reason isn't giving me the ability to restart through the UI. Can you see if you can restart it or push an empty commit?

bweissbart (changes requested):
Unfortunately, per agreements we have with clients, we cannot remove things without deprecating it for a cycle first. Can you please restore the old legacy form components and add yellow box warnings about their deprecations?

yarn.lock
```diff
@@ -3756,14 +3749,21 @@
     "@types/react" "*"
     "@types/react-native" "*"
 
-"@types/react-native@*", "@types/react-native@^0.60.22":
+"@types/react-native@*":
```
Cauldrath, line 3752: We should be able to delete this entry and yarn install again so that there's only one version of the types loaded.

Cauldrath:
Conflicts.

packages/flagship/ios/FLAGSHIP.xcodeproj/project.pbxproj
```diff
@@ -342,12 +365,18 @@
 				INFOPLIST_FILE = FLAGSHIP/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 10.3;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
```
Cauldrath, line 367: I had to change this line to:
`LD_RUNPATH_SEARCH_PATHS = "/usr/lib/swift $(inherited) @executable_path/Frameworks";`
to get it to run.

Cauldrath:
Trying this both with and without merging in the feature branch, it looks like images aren't loading on both iOS and Android.

Cauldrath:
> Unfortunately, per agreements we have with clients, we cannot remove things without deprecating it for a cycle first. Can you please restore the old legacy form components and add yellow box warnings about their deprecations?

tcomb-form-native hasn't been updated since 10-11-18 and the last supported version of RN it has listed is 0.35.0. Can we deprecate it in FS9?

packages/flagship/ios/FLAGSHIP.xcodeproj/project.pbxproj
nathan-sankbeil: Thanks for taking care of that. I was able to run iOS without it, but adding it doesn't seem to harm anything.

Cauldrath (approved):
Seems good. Pirateship runs on iOS, Android, and web. Some of the commit messages fail the linter, though, which is what is causing the Travis build to fail.

PR #1657, opened 2020-09-16
feat: replace react-native-restart with RN DevSettings API

nathan-sankbeil:
depends on: https://github.com/brandingbrand/flagship/pull/1656


Cauldrath (approved):
Seems good. I did verify that this doesn't fix the RNN iOS bug, though.

PR #1662, opened 2020-09-21
feat(fsweb): add proxy support

ooleksoloviov:
### Description 

Set up a generic application to proxy

### Proxy setup guide

1. Configure dataSource environment : 
a. Enable proxy
b. Change Network API request URL to the configured Proxy API.
```
dataSource: {
    enableProxy: true,
    ...otherCommonConfig,
    "networkClient": {
        "baseURL": "https://localhost:3000/path/to/proxy/api",
         ...otherCommonConfig,
      },
} 
```
c. Setup CORS options for proxy server (optional) 
2. Create configuration for proxy
```
module.exports = {
  changeOrigin: true
}
```
a. By default create the file config.js in proxy folder (optional in root path)


packages/fsweb/proxy/proxy.js
```diff
@@ -0,0 +1,73 @@
+const express = require('express');
+const proxy = require('http-proxy-middleware');
+const url = require('url');
+const { removeDWCookies, mutateDWCookies } = require('./lib');
```
Cauldrath, line 4: We should make the Demandware-specific code an option that you can turn on, so the proxy can support other apis.

packages/fsweb/package.json
```diff
@@ -8,7 +8,9 @@
     "build": "webpack --mode production",
     "build:size": "webpack --json | webpack-bundle-size-analyzer",
     "start": "webpack-dev-server -d --config ./webpack.config.js --mode development --inline --hot",
-    "start-prod": "webpack-dev-server -d --config ./webpack.config.js --mode production"
+    "start-prod": "webpack-dev-server -d --config ./webpack.config.js --mode production",
+    "server-dev": "node ./proxy/server.dev.js",
+    "server-prod": "node ./proxy/server.prod.js"
```
Cauldrath, line 13: These commands should also be added to pirateship, fstestproject, and `ship:` versions added to the base package. The files also need to be updated so that they will point at the env pile in the project.

packages/fsweb/proxy/server.dev.js
```diff
@@ -0,0 +1,43 @@
+const exec = require('child_process').exec;
+const path = require('path');
+const Proxy = require('./proxy');
+
+const port = process.env.PORT || 3000;
+const hostname = process.env.HOST || 'localhost';
+const envPath = process.env.ENV_PATH || '../../env';
```
Cauldrath, line 7: This seems like it needs to be: (also on prod)
`const envPath = process.env.ENV_PATH || 'env/env';`

packages/fsweb/proxy/server.prod.js
```diff
@@ -0,0 +1,29 @@
...
+const Proxy = require('./proxy');
+
+const envPath = process.env.ENV_PATH || '../../env';
+const env = require(
+  path.resolve(__dirname, '../../', envPath)
+);
+const port = process.env.PORT || 3000;
+
+if (env && env.dataSource.type === 'commercecloud') {
+  console.log('Env not found in path, ', envPath);
```
Cauldrath, line 11: This shows up every time the server starts and looks like an error message.

packages/fsweb/proxy/server.prod.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const proxy = new Proxy(env.dataSource.apiConfig).initProxy();
+
+  if (!proxy) {
+    return;
+  }
+
+  const index = path.resolve(__dirname, '..', 'web', 'build', 'index.html');
```
Cauldrath, line 19: This needs to be the compiled version, because the bundle.js file needs to be injected.

ooleksoloviov: @Cauldrath done

packages/fsweb/proxy/server.prod.js
ooleksoloviov: @Cauldrath done

packages/fsweb/proxy/server.dev.js
ooleksoloviov: @Cauldrath done

packages/fsweb/package.json
ooleksoloviov: @Cauldrath done

packages/fsweb/proxy/proxy.js
ooleksoloviov: @Cauldrath done

packages/fsweb/proxy/server.prod.js
```diff
@@ -0,0 +1,56 @@
...
+
+  if (!proxy) {
+    return;
+  }
+
+  const build = path.resolve(__dirname, rootDir, buildPath);
+
+  proxy.use(
+    express.static(build)
+  );
```
Cauldrath, line 47: This happens after the proxy's app.use, so it will never actually serve any of these files. I was able to get it to serve the files with this:
```
  const proxy = new Proxy(apiConfig, proxyConfig);
  const proxyApp = proxy.app;

  if (!proxyApp) {
    return;
  }

  const build = path.resolve(__dirname, rootDir, buildPath);

  proxyApp.use(
    express.static(build)
  );

  proxy.initProxy();
```

It might be better to pass the buildPath into initProxy instead and have it set the static middleware.

Cauldrath (changes requested):
There also currently is no way to build the web-compiled folder besides manually going into the /web folder and running the full command. We should copy the build-web yarn command (including a root-level ship: command) and buildWeb.sh file from the template into pirateship.

The dev version seems to not serve files at all. I assume the proxy is meant to be run along side the compile-web version?

Cauldrath:
Oh, also please add `packages/pirateship/web-compiled` to the .gitignore

ooleksoloviov:
@Cauldrath according to your question. We do not need to serve the dev app in express as we do this using webpack-dev-server. Also, live reload will not work if we serve the dev app with express. That is why I add the CORS configuration to make possible for the user set the dev app origin as allowed to work alongside with proxy server.

packages/fsweb/proxy/server.prod.js
ooleksoloviov: Done

Cauldrath (approved):
Seems good. I think we should run the server regardless of whether the proxy is configured, but what is here works for what was asked.

PR #1672, opened 2020-09-22
chore: reorder ci test commands

bweissbart:
This moves commitlint to the end of the commands run to validate PRs. This way we can see if the "real" tests pass first even if commitlint fails.


Cauldrath (approved):
Seems good.

PR #1680, opened 2020-09-25
feat: make permissions a soft dependency in fsengage

bweissbart:
fsengage is a dependency on a number of fs packages, and fsengage had a hard dependency on react native permissions due to its geolocation handler for the core cms provider. react-native-permissions requires native setup and exposes itself to end users, which is not ideal given that most projects that use fsengage don't need geolocation functionality. This PR makes permissions a peer dependency and updates the geolocation handler to return false if react native permissions isn't available in the project.

### DIFF
--- a/packages/fsengage/package.json
+++ b/packages/fsengage/package.json
@@ -22,7 +22,8 @@
     "url-parse": "^1.4.3"
   },
   "devDependencies": {
-    "sinon": "^8.0.2"
+    "sinon": "^8.0.2",
+    "react-native-permissions": "^2.2.0"
   },
   "peerDependencies": {
     "react-native-permissions": "^2.2.0"
--- a/packages/fsengage/package.json
+++ b/packages/fsengage/package.json
@@ -19,11 +19,13 @@
     "decimal.js": "^10.0.0",
     "react": "^16.13.1",
     "react-native": "^0.63.2",
-    "react-native-permissions": "^2.2.0",
     "url-parse": "^1.4.3"
   },
   "devDependencies": {
     "sinon": "^8.0.2"
+  },
+  "peerDependencies": {
+    "react-native-permissions": "^2.2.0"
   },
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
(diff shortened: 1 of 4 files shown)


Cauldrath:
Theoretically, you should be able to set react-native-permissions as a devDependency, as well, then `import type` so it doesn't include the import in the Javascript file and you can set appropriate types, like this:
```
import { Platform } from 'react-native';
import type { default as RNPermissions, Permission } from 'react-native-permissions';

// Geolocation is currently bundled with fsengage even though apps may not use location-based
// targeting. Because react-native-permissions requires apps to expose the permissions they use,
// this makes permissions a "soft" dependency that apps can opt into by including rn-permissions
// as a dependency. TODO - refactor fsengage so that location-based permissions are taken out of
// engage core.
let rnPermissions: typeof RNPermissions | null = null;

try {
  rnPermissions = require('react-native-permissions');
} catch (e) {
  console.warn(
    'react-native-permissions must be added to your project'
    + ' to enable granular geolocation in fsengage'
  );
}

function getPermissionToCheck(rnPermissions: typeof RNPermissions): Permission {
  return Platform.OS === 'ios' ?
    rnPermissions.PERMISSIONS.IOS.LOCATION_WHEN_IN_USE
      : rnPermissions.PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION;
}

export async function isGeolocationAllowed(): Promise<boolean> {
  const permissions = rnPermissions;
  if (permissions) {
    const permissionToCheck = getPermissionToCheck(permissions);

    return permissions.check(permissionToCheck)
      .then(status => status === permissions.RESULTS.GRANTED)
      .catch((error: Error) => {
        if (__DEV__) {
          console.log(
            `%cLocator\n%c Function: isGeolocationAllowed\n Error: `,
            'color: blue',
            'color: grey',
            error
          );
        }

        throw error;
      });
  }

  return Promise.resolve(false);
}

export async function requestGeolocationPermission(): Promise<boolean> {
  const permissions = rnPermissions;
  if (permissions) {
    const permissionToCheck = getPermissionToCheck(permissions);
    return permissions.request(permissionToCheck)
      .then(status => status === permissions.RESULTS.GRANTED);
  }

  return Promise.resolve(false);
}
```

bweissbart:
@Cauldrath - updated. Thanks for the suggestion! Always good to learn something new.

Cauldrath (approved):
Seems good.

PR #1681, opened 2020-09-25
fix: audit usenativedriver settings for animations

bweissbart:
Per the React Native docs, using the native driver for animations will only work for non-layout transforms like opacity and rotation. RN appears to have silently ignored invalid transforms in the past, but in RN 63 it displays a promiment warning. This updates all invalid transforms using native drivers to set useNativeDriver to false.

Reference which is from 2017 but still appears to be valid per my research: https://reactnative.dev/blog/2017/02/14/using-native-driver-for-animated.html#caveats


Cauldrath (approved):
Seems good.

PR #1682, opened 2020-09-25
fix(fsengagement): address cannot set state on unmounted component error

bweissbart:
This addresses a warning regarding updating state on an unmounted component in EngagementComp. It adds a simple boolean keeping track of whether the component is currently mounted so that async operations can check before running.


Cauldrath (approved):
Seems good.

PR #1688, opened 2020-09-29
fix(flagship): Update to async-storage 2

Cauldrath:
async-storage 1 doesn't support web and the latest version of react-native-web doesn't have async-storage in it anymore.


Cauldrath:
1.9.0 has web support, so keeping this work for when 2.0.0 is fully released. 1.9.0 used here:
https://github.com/brandingbrand/flagship/pull/1689

bweissbart:
react-native-community/react-native-storage has transferred ownership to react-native-async-storage/async-storage, and as best as I can tell they aren't currently developing or releasing the 2.0 branch that this PR builds against.

PR #1689, opened 2020-09-29
fix(fsweb): Update to newer version of async-storage

Cauldrath:
Web support was added in 1.9.0. This is updating to 1.12.0 in the yarn.lock, though.


bweissbart (approved):
👍

PR #1690, opened 2020-09-29
feat: replaced tcomb-form-native with bb fork

bweissbart:
tcomb-form-native hasn't been updated in several years and therefore still had dependencies on React Native that had been moved to separate packages as part of the RN lean core project.

We created a Branding Brand fork, @brandingbrand/tcomb-form-native, that addresses these issues. This updates Flagship to use the forked version.


Cauldrath (approved):
Seems good.

PR #1692, opened 2020-09-30
chore: upgrade babel-plugin-react-native-web to 0.13.14

bweissbart:


bweissbart:
@Cauldrath - do you mind taking a look at this? I think it's something we'll probably want to go with our RN web upgrade.

Cauldrath (approved):
Seems good. Storybook, compile-web, and ssr all work.

PR #1696, opened 2020-10-05
feat(fsapp): Add SSR support

Cauldrath:
Currently just populates the initial state of the redux container, but has support for setting variables between client and server that can be used for other components. Also still generates assets in the SSR build that aren't actually being used, but I was getting compilation errors when I tried to remove them, so I'll be addressing that in a later PR. I'd also like to eventually unify the web and native config files more so that having separate ones won't be necessary in the future.

TEST
-----------------------
- yarn
- yarn ship:init-web
- yarn ship:build-web (builds client-side code)
- yarn ship:build-ssr (builds server-side code)
- yarn ship:start-server-ssr


bweissbart (approved):
💯

PR #1702, opened 2020-10-06
chore(fsengage): document new gaqueryparams added to engage

bweissbart:


packages/fsengage/README.md
```diff
@@ -462,6 +475,24 @@ analytics.lifecycle.suspend(): void;
...
+The Google Analytics provider has support for specifying custom parameters by way of setting  `gaQueryParams` in the payload. For example:
+
+```javascript
+analytics.impression.product(
+  component: React.Component | string,
+  properties: {
+    identifier: 'abc123',
+    name: 'Large Blue Pants',
+    gaQueryParams: {
+      coupon: 'SALE45'
```
nathan-sankbeil, line 491: This isn't a good example since it's not valid. We should include a link to the reference guide to see what query string parameters can be used.

https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters

nathan-sankbeil (changes requested):

packages/fsengage/README.md
bweissbart: My bad, I had a whole other commit that I forgot to push up. This should look better now. Damn end of day work...

nathan-sankbeil (approved):

PR #1715, opened 2020-10-08
feat(fscomponents): Update SearchBar and SearchScreen

Cauldrath:


nathan-sankbeil (changes requested):
Can you update the searchBarProps styling in the SearchScreen story? It has a double border. You didn't introduce the weird behavior, but we should clean it.

image

nathan-sankbeil (approved):

PR #1722, opened 2020-10-09
feat(fscomponents): add NestingButtons component

nathan-sankbeil:
Accepts an array of button props and generates buttons based on those. If the number of buttons
exceeds a configurable threshold, the buttons will be hidden behind a modal.

nesting-buttons


Cauldrath (approved):
Seems good.

PR #1723, opened 2020-10-09
feat(fscomponents): Update FilterListDrilldown

Cauldrath:
Support for a close and apply button and setting the text of the refine button, as well as some other minor features. Also added a story for it.

PR #1724, opened 2020-10-09
feat(fsproductindex): Updated ProductIndex

Cauldrath:
Builds on https://github.com/brandingbrand/flagship/pull/1723

Added functions for handling when filters are applied or reset, setting total item count, and rendering a ghost screen, as well as toggles for cancel and close. Also some handling for sorts and removing refinements from the list.


nathan-sankbeil (approved):
Seems good.

PR #1735, opened 2020-10-14
feat(fscomponents): Update StepIndicator

Cauldrath:
Can toggle between the old and new versions by setting "line" to true.


nathan-sankbeil (approved):
Seems good.

PR #1738, opened 2020-10-15
fix(flagship): update RNFirebase pod

griffithcaleb:
**Description**
- the `react-native-firebase` module was adding `Firebase/Core 5.1.2` to the Podfile
- this was causing a build error because the react-native 60+ is dependent on iOS SDK 6.13+ 
- updating to add 6.13 to the Podfile and its dependency `pod 'GoogleAppMeasurement', '6.1.6'`
 
from https://www.npmjs.com/package/react-native-firebase
image


Cauldrath (approved):
Seems good.

PR #1749, opened 2020-10-19
fix(flagship): Update storybook and rn-device-info

Cauldrath:
Also sets storybook action depth to 1 and rebuilds yarn.lock from scratch to avoid module duplication. Fixes some actions throwing errors instead of displaying in the storybook and debugger not working on Android.

Tested typedoc, storybook, web, android, and ios.


.storybook/main.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  stories: [
+    '../packages/**/*.@(story|stories).@(ts|tsx)'
+  ],
+  addons: [
+    '@storybook/addon-actions/register',
+    '@storybook/addon-knobs/register'
```
nathan-sankbeil, line 7: Two things here:

- I think you missed the viewport add-on. It was previously defined in the [addons.js file](https://github.com/brandingbrand/flagship/blob/develop/.storybook/addons.js#L3) and is missing from this array.
- Can you change the order of these so that the knobs add-on is first? The load order seems to impact the tab order, so the actions tab is initially shown now. Having the knobs tab initially shown seems more useful. (This may be configurable through some other option too.)

nathan-sankbeil (changes requested):
Do we explicitly need to upgrade react-native-device-info since it's technically a breaking change? In testing, it seemed like simply using the async version of `getUserAgent` was sufficient.

nathan-sankbeil (approved):

PR #1758, opened 2020-10-22
chore: merge develop into master

bweissbart:


Cauldrath (approved):
Seeems good. There are no commits on master that aren't on develop, so this should just make master the same as develop.

PR #1759, opened 2020-10-22
feat(fscomponents): Updates to Swatches

Cauldrath:
Updated for new version of OptionValue, which has swatch data in it. Deprecated color and image, which are now redundant. Also added event override for clicking the plus.


nathan-sankbeil (changes requested):
- Selecting a swatch no longer updates the UI
swatches

- Styling for the show more button is wonky.
image

nathan-sankbeil (approved):

PR #1765, opened 2020-10-27
fix: fix android release variant compilation

nathan-sankbeil:
- adds path function to get debug android src folder
- adds init step to rename debug src folder structure to match package
- fixes flipper package name
- moves flipper to only be included in debug


Cauldrath (approved):
Seems good.

PR #1770, opened 2020-10-29
fix(flagship): made rn camera patch script work again

bweissbart:
react-native-camera needs a missingDimensionStrategy line added to build.gradle to build properly. Our script to patch this was looking to do a string replace on a value that no longer exists in the file.


Cauldrath (approved):
Seems good.

PR #1785, opened 2020-11-05
fix(fsweb): Remove svgs module

Cauldrath:
react-native-svg has web support now, so a separate module isn't needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,6 @@
     "react-native-web": "^0.13.12",
     "react-native-web-modal": "^1.0.1",
     "rimraf": "^3.0.0",
-    "svgs": "^4.1.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -98,6 +98,7 @@
     "react-native-web": "^0.13.12",
     "react-native-web-modal": "^1.0.1",
     "rimraf": "^3.0.0",
+    "svgs": "^4.1.0",
```
nathan-sankbeil, line 101: Since we're removing `svgs`, do we need to add this?

Cauldrath: Of course not.

nathan-sankbeil (approved):

PR #1787, opened 2020-11-06
feat(flagship): remove flipper support

bweissbart:
Flipper, while promising, is causing conflicts with many native libraries, requiring developers to manually remove it from their projects. This removes Flipper completely for now; hopefully React Native can find an easier integration path going forward.


Cauldrath (approved):
Seems good.

Issue #1788, opened 2020-11-09
fscomponents: NestingButtons - Cannot close modal when button is pressed


wSedlacek:
https://github.com/brandingbrand/flagship/blob/eecc4eaab727f796e5f9ad718464fcfe80cc8f71/packages/fscomponents/src/components/NestingButtons.tsx#L73-L79

At this time it is impossible to trigger the `ModalHalfScreen` outside of the component which is resulting in some strange behavior in Plan A

PR #1789, opened 2020-11-09
feat(fscomponents): triggered modal closing

wSedlacek:
## Description

This adds a new option to the `buttonProps` of `NestedButtons` that will allow for a button to trigger the modal to close.
As it stands, it is not possible to close the modal from outside of NestedButtons which results in strange behavior.

## Dev Notes

At this time this option defaults to false to maintain the same behavior as already exist but simply adding a new option to change the behavior.
In a future release I would like to change this option to default to true as I believe that behavior is more expected but would be a breaking change.

fixes: #1788


bweissbart:
Spoke with @wSedlacek and he believes these changes have been superseded by how modals are handled in FS11. Closing.

PR #1805, opened 2020-11-25
feat: add ExtractPropertyNames type

nathan-sankbeil:
Extracts the property names from a given object whose values are assignable to a given type

Example:

```
interface MyObject {
  a: string;
  b: string;
  c: () => void;
  d: number;
}

type StringProperties = ExtractPropertyNames<MyObject, string>;
//                  ^ = type StringProperties = 'a' | 'b'

type NumberProperties = ExtractPropertyNames<MyObject, number>;
//                  ^ = type NumberProperties = 'd'
```

### DIFF
--- a/packages/fsfoundation/src/ExtractPropertyNames.ts
+++ b/packages/fsfoundation/src/ExtractPropertyNames.ts
@@ -1,0 +1,20 @@
+/**
+ * Extracts the property names from a given object whose values are assignable to a given type
+ *
+ * Example:
+ * interface MyObject {
+ *  a: string;
+ *  b: string;
+ *  c: () => void;
+ *  d: number;
+ * }
+ * type StringProperties = ExtractPropertyNames<MyObject, string>;
+ *                     ^ = type StringProperties = 'a' | 'b'
+ *
+ * type NumberProperties = ExtractPropertyNames<MyObject, number>;
+ *                     ^ = type NumberProperties = 'd'
+ *
+ */
+export type ExtractPropertyNames<T, S> = {
+  [K in keyof T]: T[K] extends S ? K : never;
+}[keyof T];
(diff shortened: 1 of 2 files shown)


packages/fsfoundation/src/ExtractPropertyNames.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Extracts the property names from a given object whose values are assignable to a given type
+ *
+ * Example:
```
wSedlacek, line 4: Shouldn't this be `@example` to follow the jsdocs spec?
I believe using `@example` would also provide the code snippet with syntax highlighting in tools like VS Code

packages/fsfoundation/src/ExtractPropertyNames.ts
```diff
@@ -0,0 +1,20 @@
...
+ * interface MyObject {
+ *  a: string;
+ *  b: string;
+ *  c: () => void;
+ *  d: number;
```
wSedlacek, line 9: Shouldn't these have one more space to be two spaces from where `interface` is indented to?

Cauldrath:
This has an issue if you pass it a type with optional keys. It winds up shoving `undefined` into the list. It doesn't have the same issue if you use `| undefined` instead of the question mark, though.

```
interface TestType {
  required: string;
  optionalWithOr: string | undefined;
}

interface TestTypeWithOptional extends TestType {
  optional?: string;
}

type TestKeys = ExtractPropertyNames<TestType, string>;
// "required"
type TestWithUndefined = ExtractPropertyNames<TestType, string | undefined>;
// "required" | "optionalWithOr"

type TestKeysWithOptional = ExtractPropertyNames<TestType, string>;
// "required" | undefined
type TestWithUndefinedWithOptional = ExtractPropertyNames<TestType, string | undefined>;
// "optional" | "required" | "optionalWithOr" | undefined
```

Wrapping it in NonNullable to get rid of undefined seems to work as a quick fix:
```
export type ExtractPropertyNames<T, S> = NonNullable<{
  [K in keyof T]: T[K] extends S ? K : never;
}[keyof T]>;
```

bweissbart:
Closing per comments.

PR #1819, opened 2020-11-30
fix(fsapp): Fix screen name handling for mangled js

Cauldrath:
This was preventing production web from working.


bweissbart:
Closing due to new direction for screen linking.

PR #1825, opened 2020-12-02
feat(flagship): Support for gifs and WebP in Android

Cauldrath:
Uses code from:
https://reactnative.dev/docs/0.62/image#gif-and-webp-support-on-android

By default, apps will now include support for both formats. Add build.gifSupport=false to turn off gif support and/or build.webPSupport=false to turn off WebP support.


bweissbart (approved):
👍

PR #1829, opened 2020-12-04
fix(ssr): fix server side rendering of the $1 instances

buticm:
## Description

BREAKING CHANGE:
Fix the rendering on the server side when the body or variables contain $1 instances.

## Test

- set any variable in the initial data function that contain a string with `$1`, e.g. `test = 'the price is $100 - $125'`
- render some text element that contain a string `$1`, e.g. `<Text>the price is $100 - $125</Text>
- load the page and view source.

ER: the variables are rendered as JSON correctly, the page body is rendered correctly.


Cauldrath (approved):
Seems good.

PR #1830, opened 2020-12-04
feat(category): allow fetchCategory query to have levels parameter

buticm:
## Description

RELATED PR:
https://github.com/brandingbrand/submarine/pull/257

The query for fetchCategory function can specify 'levels' parameter to get more levels of categories.
This allows to avoid additional requests in some cases when we need to get more than 2 levels of categories,
that are provided by default.


Cauldrath (approved):
Seems good.

PR #1860, opened 2020-12-16
fix(flagship): set proper device targets in ios template

bweissbart:
Flagship only supports iOS on the Apple side, so we should only be targeting TARGETED_DEVICE_FAMILY="1" in the pbxproj file. We were incorrectly targeting 1,2 which corresponds to a universal app.

I also added a unit test to try to avoid any regressions in the future.


nathan-sankbeil (approved):
🔥

PR #1865, opened 2020-12-29
fix(flagship): Remove animated image patch

Cauldrath:
React-native has been updated with its own fix for this issue, with a slight difference:
```
if (_currentFrame) {
    layer.contentsScale = self.animatedImageScale;
    layer.contents = (__bridge id)_currentFrame.CGImage;
  } else {
    [super displayLayer:layer];
  }
```
vs
```
  if (_currentFrame) {
    layer.contentsScale = self.animatedImageScale;
    layer.contents = (__bridge id)_currentFrame.CGImage;
  }
  [super displayLayer:layer];
```

The version without the else was breaking gifs on iOS 12, so this removes the patch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "react": "^16.13.1",
     "react-art": "^16.4.0",
     "react-dom": "^16.9.0",
-    "react-native": "^0.63.2",
+    "react-native": "^0.63.3",
     "react-native-navigation": "^6.0.0",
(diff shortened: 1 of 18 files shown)


bweissbart:
The relevant portion of this PR, which doesn't upgrade RN, was merged here: https://github.com/brandingbrand/flagship/pull/1871

PR #1872, opened 2021-01-05
fix(fsapp): fix typing for app router

andyschwob:
In my original PR there were a lot of `any`s so I added some typing and also moved some types from the flagship-demo repo in here where it should be.

### DIFF
--- a/packages/fsapp/src/lib/app-router.ts
+++ b/packages/fsapp/src/lib/app-router.ts
@@ -1,5 +1,5 @@
 // tslint:disable:cyclomatic-complexity
-import FSNetwork, { AxiosResponse } from '@brandingbrand/fsnetwork';
+import FSNetwork from '@brandingbrand/fsnetwork';
 import { LayoutComponent } from 'react-native-navigation';
 import Navigator from './nav-wrapper';
 import pathToRegexp, { Key } from 'path-to-regexp';
@@ -12,6 +12,11 @@ import {
   PublishedPage,
   RouterConfig
 } from '../types';
+
+export interface TestType {
+  [key: string]: string;
+}
+
 
 export default class AppRouter {
   cmsToken: string = '';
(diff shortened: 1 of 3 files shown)


packages/fsapp/src/lib/app-router.ts
```diff
@@ -302,7 +303,10 @@ export default class AppRouter {
         }
 
         return customRoutes.filter((page: PublishedPage) => page.path !== null)
-          .reduce((ret: any, page: any) => {
+          .reduce((ret: any, page: PublishedPage) => {
```
wSedlacek, line 306: Still noticing an `any` here that relates back to line 263 `async fetchPageRoutes(): Promise<any>`

From what I can tell this should be typed `Recored<string, Pick<PublishedPage , 'id' | 'title'>`

packages/fsapp/src/lib/app-router.ts
```diff
@@ -273,12 +274,12 @@ export default class AppRouter {
           Authorization: `Bearer ${this.cmsToken}`
         }
         // tslint:disable-next-line: cyclomatic-complexity
-      }).then(async (pageData: any) => {
+      }).then(async (pageData: AxiosResponse) => {
```
wSedlacek, line 277: I would personally prefer that types that can be inferred, like callback types from promises, to be inferred. This tends to feel more natural for me and allows `.get()` to use a generic to control this type.

wSedlacek (changes requested):
So much easier to read with the types being defined! There were a few that we still haven't defined types for yet but this is so much better.

packages/fsapp/src/lib/app-router.ts
andyschwob: @wSedlacek changing this now, but i believe it should be:
`Record<string, Omit<PublishedPage, 'id' | 'path'>>,`

wSedlacek: `Omit` would exclude everything but the `id` and `path` where as on lines 310-312 it seems that we are defining `id` and `title`. That said I do see that `id`is renamed to `pageId` on those lines so the type I provided wasn't quite right on that aspect.

Sounds like you are on the right track though.

wSedlacek (approved):
LGTM

wSedlacek:
@bweissbart Everything looks good here, feel free to merge if you are happy with it.

bweissbart (approved):

PR #1905, opened 2021-01-27
fix(fsapp): linked-app fixes after debugging

wSedlacek:
This includes many fixes to `linked-app` after debugging within the web environment of `flagship-demo`.
Some of these fixes are simple things like binding methods, or dependency version issues.
Others are like missed variables that were needed like `matchedMath`.

This also includes an initial implementation of the modal system but disabled from the public API due to issues on the web platform. A future change will enable these APIs and fix the issues on web.


bweissbart:
@wSedlacek - tsc error:

```
src/linked-app/History/types.ts(10,1): error TS6133: 'ModalComponentType' is declared but its value is never read.
```

bweissbart (approved):

PR #1915, opened 2021-02-03
[FS-2035] :  Modify  XCode project

bhushankini:
Use plutil to make changes to Project

### DIFF
--- a/packages/flagship/src/lib/ios.ts
+++ b/packages/flagship/src/lib/ios.ts
@@ -399,16 +399,8 @@ export function iosExtensions(configuration: Config, version: string): void {
     );
 
     // Update Extension PList
-    fs.update(
-      extPlistPath,
-      /\<key\>CFBundleShortVersionString\<\/key\>[\n\r\s]+\<string\>[^\s]+<\/string\>/,
-      `<key>CFBundleShortVersionString</key>\n\t<string>${version}</string>`
-    );
-    fs.update(
-      extPlistPath,
-      /\<key\>CFBundleVersion\<\/key\>[\n\r\s]+\<string\>[^\s]+<\/string\>/,
-      `<key>CFBundleVersion</key>\n\t<string>${bundleVersion}</string>`
-    );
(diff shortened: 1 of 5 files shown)


packages/flagship/src/lib/ios.ts
```diff
@@ -21,11 +24,8 @@ export function bundleId(configuration: Config): void {
 
   helpers.logInfo(`updating iOS bundle id`);
 
-  fs.update(
-    path.ios.infoPlistPath(configuration),
-    /<key>CFBundleIdentifier<\/key>\s+<string>[^<]+<\/string>/,
-    `<key>CFBundleIdentifier</key><string>${bundleId}</string>`
-  );
+  exec(`plutil -replace CFBundleIdentifier -string ${bundleId} ${path.ios.infoPlistPath(configuration)}`);
```
skyeckstrom, line 27: Instead of using `execSync` and passing unsantizied inputs, we should use `execFileSync` so the `bundleId` can be escaped properly. The risk with this approach is a malicious Flagship project could pass in shell code in the bundleId that would be blindly executed.

Change this to something like
```typescript
execFileSync('plutil', ['-replace', 'CFBundleIdentifier', '-string', bundleId, path.ios.infoPlistPath(configuration)])
```

packages/pirateship/env/common.js
```diff
@@ -46,5 +46,6 @@ module.exports = {
     ios: 'com.brandingbrand.reactnative.pirateship',
   },
   associatedDomains: [],
-  targetedDevices: 'Universal'
+  targetedDevices: 'Universal',
+  UIBackgroundModes: ['remote-notification']
```
skyeckstrom, line 50: Can you remove this unrelated change?

skyeckstrom (changes requested):

packages/flagship/src/lib/ios.ts
bhushankini: done

packages/pirateship/env/common.js
bhushankini: done

PR #1931, opened 2021-02-15
chore: update serializable components to use fsbeta navigator

andyschwob:
@bweissbart @wSedlacek 

There is a matching layoutbuilder pr to this: https://github.com/brandingbrand/submarine/pull/346

This pulls in the FSBeta navigator for our serializable components.  If written it to the default action is to use `navigator.open()`, but you can also pass in `onPress` to override this action.  For example, in NA we can use `withPropTransformer` to use a custom onPress to override.  

Also, I did check and confirm the default Navigator context for FSBeta works fine in NA...this is the case where the Navigator context does not exist so it uses dummy.navigator in FSBeta.  The `onPress`/`handlePress` doesn't even get called on the canvas, if you switch on "preview" mode it does indeed get called but it simply throws an error in the console as intended and does not crash.


packages/fscomponents/src/components/serializable/SerializableImage.tsx
```diff
@@ -21,19 +22,28 @@ export interface SerializableImageProps
...
+      }
+    };
+
+    if (!(href || onPress)) {
       return <Image {...props} style={[self, host]} />;
     }
 
     return (
-      <TouchableOpacity style={host} onPress={onPress}>
+      <TouchableOpacity style={host} onPress={handlePress(href)}>
```
wSedlacek, line 46: We should just be able to use `href` directly from props rather then using a HOF here which should improve performance a bit as TouchableOpacity will no longer have it's onPress changing every render.

packages/fscomponents/src/components/serializable/SerializableImagePlaceholder.tsx
```diff
@@ -1,15 +1,25 @@
...
-    return <TouchableOpacity onPress={onPress}>{img}</TouchableOpacity>;
+    const handlePress = (href?: string) => () => {
+      if (onPress) {
+        onPress(href);
+      } else if (href) {
+        navigator.open(href);
+      }
+    };
+
+    return <TouchableOpacity onPress={handlePress(href)}>{img}</TouchableOpacity>;
```
wSedlacek, line 23: Same for here, a HOF is unnecessary

packages/fscomponents/src/components/serializable/SerializableText.tsx
```diff
@@ -23,12 +24,32 @@ export interface SerializableTextProps
...
+
+    const handlePress = (href?: string) => () => {
+      if (href) {
+        navigator.open(href);
+      }
+    };
+
+    if (href) {
+      return (
+        <TouchableOpacity onPress={handlePress(href)}>
```
wSedlacek, line 42: And here as well

packages/fscomponents/src/components/serializable/SerializableButton.tsx
```diff
@@ -13,6 +13,7 @@ import { ButtonProps } from '../Button';
 import { border, palette as defaultPalette } from '../../styles/variables';
 import { style as S, stylesSize, stylesTextSize } from '../../styles/Button';
 import { extractHostStyles } from '../../lib/style';
+import { useNavigator } from '@brandingbrand/fsapp';
```
wSedlacek, line 16: I would like to keep the props sorted in the following order.
```
type imports

react / react-native / redux

lodash / other utils / @brandingbrand

local / internal modules.
```

So for this import I would like for it to be between lines 10 and 12 with new lines on either side.

```
} from 'react-native';

import { useNavigator } from '@brandingbrand/fsapp';

import { ButtonProps } from '../Button';
```

wSedlacek (approved):
Overall this is a good change and what I had in mind, that said,

There is still room for improvement as there always is, most notably for performance we should avoid using higher order functions in our returned JSX as every rendering of the JSX would invoke the function and create a new function definition which would be a different value from the previous rendering thus would be considered a props change.

Similarly when defining call backs it would be best to use `useCallback` so that the function definition is memoized.

Resources: [jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md), [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback)

andyschwob:
@wSedlacek i've addressed the issues you mentioned

wSedlacek (approved):
Looks great!

bweissbart (approved):

PR #1951, opened 2021-02-26
fix(fscommerce): session manager

buticm:
## Description

BREAKING CHANGE:
- allow only one invocation of the session manager `get` method at a time.
- provide types for the `fetchCart` method to allow only one invocation of the method at a time.

## Test

Test together with the related PR.
https://github.com/brandingbrand/submarine/pull/365

### DIFF
--- a/packages/fscommerce/src/Commerce/sessions/CommerceCookieSessionManager.web.ts
+++ b/packages/fscommerce/src/Commerce/sessions/CommerceCookieSessionManager.web.ts
@@ -23,6 +23,15 @@ export default class CommerceCookieSessionManager extends CommerceSessionManager
   }
 
   /**
+   * Provides cookies to decide if the session can be converted to JWT
+   *
+   * @returns {Promise.<string | null>} A Promise representing cookies string
+   */
+  async getCookies(): Promise<string | null> {
+    return document.cookie;
+  }
+
+  /**
    * Returns current token for the session.
    *
    * Only one invocation at a time is allowed. If it's called again
@@ -32,9 +41,9 @@ export default class CommerceCookieSessionManager extends CommerceSessionManager
    *
    * The token is refreshed automatically upon expiration if `refreshToken` method was provided.
    *
-   * If no token exist it will attempt to convert current session into JWT,
-   * but only if the cookie `dwanonymous_*` is present.
-   * Otherwise, it will create new token for the guest user.
+   * If no token exist it will attempt to convert current session into JWT.
+   *
+   * If we could not convert current session, it will create new token for the guest user.
    *
    * @returns {Promise.<SessionToken | null>} A Promise representing token
    */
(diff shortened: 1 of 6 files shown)


notajvento:
@bweissbart / @nathan-sankbeil 
if we could you guys to take a look here it would be a huge help. This is pretty high priority

packages/fscommerce/src/Commerce/types/Cart.ts
```diff
@@ -37,6 +37,12 @@ export interface Cart<T extends CartItem = CartItem> {
    */
   items: T[];
 
+  /**
+   * An array of products with detailed info for variants in the cart, their master
+   * products and bonus products in the cart.
+   */
+  product_details?: Product[];
```
nathan-sankbeil, line 44: All of our other property names are camelCased, so this should be too.

packages/fscommerce/src/Commerce/sessions/CommerceCookieSessionManager.web.ts
```diff
@@ -16,40 +22,108 @@ export default class CommerceCookieSessionManager extends CommerceSessionManager
...
+          /* let it fail sliently */
+        }
+      }
+
+      // got a token - store it
+      if (token) {
+        await this.set(token);
+
+        return token;
+      }
```
nathan-sankbeil, line 101: I'm not entirely clear on the purpose of restoreCookies, but can we simplify this by doing the cookie restore first? eg.

```ts
// ...
      if (this.options.restoreCookies) {
        try {
          await this.options.restoreCookies();
        } catch (e) {
          /* let it fail sliently */
        }
      }

      // no token - get a token from the session
      const sessionCookiesPresent = document.cookie?.includes('dwanonymous_');
      if (sessionCookiesPresent) {
        try {
          token = await this.options.sessionCookiesToToken();
          // got a token - store it
          if (token) {
            await this.set(token);
            return token;
          }
        } catch (e) {
          /* let it fail sliently */
        }
      }
// ...
```

nathan-sankbeil (changes requested):

packages/fscommerce/src/Commerce/sessions/CommerceCookieSessionManager.web.ts
```diff
@@ -16,40 +22,108 @@ export default class CommerceCookieSessionManager extends CommerceSessionManager
...
 
+      // got old token
       if (token) {
         return token;
       }
-    }
 
-    return this.options.createGuestToken();
+      // no token - get a token from the session
+      const sessionCookiesPresent = document.cookie?.includes('dwanonymous_');
```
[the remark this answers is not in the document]
buticm: @nathan-sankbeil,
I created new method that will provide cookies to the user, so this implementation will be user specific.
Using cookies is optional, so the implementation of that method is optional.
By default the method will return null.
For web implementation it will return current cookie string.

packages/fscommerce/src/Commerce/types/Cart.ts
buticm: @nathan-sankbeil updated

packages/fscommerce/src/Commerce/sessions/CommerceCookieSessionManager.web.ts
buticm: @nathan-sankbeil,
I'm not aware of the users of the method `restoreCookies` and possible implementations.
I didn't touch current implementation and wary of doing that. I don't want to introduce unnecessary breaking changes.

buticm:
@nathan-sankbeil, ready for review

nathan-sankbeil (approved):

PR #1960, opened 2021-03-05
fix: add stack id to enable tab switching on native

andyschwob:
@bweissbart debugging session with @wSedlacek led to the need to include the root stack id to enable switching tabs via `navigator.open` on native


wSedlacek (approved):

andyschwob:
I've updated with the componentId fix @bweissbart @wSedlacek

bweissbart (approved):

PR #1972, opened 2021-03-15
feat: useButtonEffect

wSedlacek:
## Description

This does three things.

a) Fixes the memory leak with `useButtons()` so that subscriptions are properly disposed of
b) Provides a new API which simplifies the process of declaring button actions via `useButtonEffect()`

```ts
useButtonEffect(() => {
  // Do something when a given button is pressed on this screen
}, [buttonId, componentId]);

// OR

useButtonEffect(() => {
  // Do something when a given button is pressed on any screen
}, [buttonId]);
```

c) Disables `leftButtons` on tab route for consistency


packages/fsapp/src/beta-app/router/context/button-context.tsx
```diff
@@ -1,42 +1,65 @@
...
+  callback: () => void,
+  ids: [buttonId: string] | [buttonId: string, componentId: string]
+) => {
+  const buttons = useButtons();
+
+  useEffect(() => {
+    if (ids.length === 1) {
+      return buttons.onPress(...ids, callback);
+    } else {
+      return buttons.onPress(...ids, callback);
```
andyschwob, line 62: just out of curiosity, is there a reason for having the exact same code in both the `if` and `else`?

wSedlacek: This is for TypeScript. Because `onPress()` has an overload TypeScript needs to know which type to check against at design time thus narrowing down the type and checking each individually is required to keep TypeScript happy.

Maybe in a future version of TypeScript it will be able to check multiple overloads at once.

andyschwob (approved):

andyschwob (approved):

PR #1988, opened 2021-03-24
fix: use tab id instead of title for tabAffinity

andyschwob:
@bweissbart @wSedlacek 
- change type `Tab` to be a tab object only (not a string)
- set tabAffinity to be the id of the tab instead of the title

### DIFF
--- a/packages/fsapp/src/beta-app/router/history/utils.base.ts
+++ b/packages/fsapp/src/beta-app/router/history/utils.base.ts
@@ -66,7 +66,7 @@ const matchPath = (path: string | undefined, route: Route) => {
 
 const buildMatcher = async (
   route: Route,
-  tab: Tab = '',
+  tab: Tab = { id: '' },
   prefix = ''
 ): Promise<
   (readonly [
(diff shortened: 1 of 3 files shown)


packages/fsapp/src/beta-app/router/history/utils.base.ts
```diff
@@ -66,7 +66,7 @@ const matchPath = (path: string | undefined, route: Route) => {
 
 const buildMatcher = async (
   route: Route,
-  tab: Tab = '',
+  tab: Tab = { id: '' },
```
wSedlacek, line 69: This should probably just be nullable with no default and use optional chaining on line 85.

A `tabAffinity` of `''` doesn't seem particularly useful or semantic.

wSedlacek (approved):
The changes look good, I did have one idea after reviewing this that you can play around with if you'd like.

andyschwob:
@wSedlacek fair enough, i was just matching the old behavior.  It's updated now.

wSedlacek (approved):
LGTM

PR #2004, opened 2021-04-05
fix: fix external route paths

andyschwob:
@bweissbart @wSedlacek 

- remove preceding slash from external route after removing tab path
- prevent route id from being registered twice


wSedlacek (approved):
LGTM

PR #2034, opened 2021-04-16
fix: apply status bar styles to Route Collection

andyschwob:
addendum to previous pr to apply status bar styles to Route Collection @wSedlacek @bweissbart


andyschwob:
@bweissbart @wSedlacek I've updated this and tested and this correctly applies the top bar styles and status bar styles.  They are applied on a route by route basis and not applied to the Route Collection.  

Example:
```
  {
    initialPath: '', // or 'discover' or 'home' or whatever we want the Discover tab route to be
    tab: Discover,
    exact: true,
    children: []
  },
  {
    initialPath: 'shop',
    tab: Shop,
    children: [
      {
        path: 'shop',
        title: TITLE,
        exact: true,
        loadComponent: () => import('./screens/Shop').then(({ Shop }) => Shop),
        topBarStyle: topBarHidden,
        statusBarStyle: {
          style: 'light',
          backgroundColor: 'black'
        }
      },
```

PR #2036, opened 2021-04-19
Feature/enable redux middleware

dtwiers:
Enables custom redux middleware to be inserted on a per-app basis.


dtwiers:
Removing to clean up git history.

PR #2037, opened 2021-04-21
fix(fsapp): allow native to have multiple tabs with same screen

andyschwob:
@bweissbart @wSedlacek 

This pr is extended from https://github.com/brandingbrand/flagship/pull/2034 so we would either need to merge that first or could just merge this and ill close the other one.  

Thanks to will for walking through this one with me, this will fix the issue of not being able to display multiple screens (with different route data) on the initialPath of multiple tabs.

For example, previously if you tried to display a Builder screen on the `/` route and a different Builder screen on the `/shop` route it would show the same data on each.


wSedlacek (approved):
LGTM

PR #2049, opened 2021-04-27
fix: REL-2442 Handling CTA button with no action and url to avoid app crash

SaurabhJ93:
Ticket: https://brandingbrand.atlassian.net/browse/REL-2442

Description:
- Added check for `handleStoryAction` function so as to handle gracefully CTA calls of stories without url and action.

Test:
- Login to app implementing fsengagement and EngagementComp.
- Navigate to a story implementing EngagementComp having CTA without any action and url.
- ER: Clicking on CTA does not crash the app and goes into `handleActionNoStory` flow.


packages/fsengagement/src/inboxblocks/CTABlock.tsx
```diff
@@ -70,6 +70,7 @@ export default class CTABlock extends Component<CTABlockProps> {
     cardPosition: PropTypes.number
   };
 
+  // // tslint:disable-next-line:cyclomatic-complexity
```
bweissbart, line 73: tslint isn't picking this up, perhaps because of the `// //`

```
ERROR: /home/travis/build/brandingbrand/flagship/packages/fsengagement/src/inboxblocks/CTABlock.tsx:74:27 - The function has a cyclomatic complexity of 11 which is higher than the threshold of 10
```

bweissbart (changes requested):

packages/fsengagement/src/inboxblocks/CTABlock.tsx
wSedlacek:
```suggestion
  // tslint:disable-next-line:cyclomatic-complexity
```

PR #2059, opened 2021-05-04
fix(fsapp): adjust findRoute to handle RouteCollection types

andyschwob:
Two fixes:
- fixes assigning the correct tabAffinity of external routes due to incorrectly appending the path of RouteCollection routes
- add `statusBarStyle` to type ExternalRoute so we can customize the status bar


packages/fsapp/src/beta-app/router/utils.ts
```diff
@@ -24,7 +24,9 @@ export const resolveRoutes = async ({
   const getChildPath = (route: Route | RouteCollection) => (
     'initialPath' in route ? route.initialPath : route.path
   );
-
+  const getPrefixedPath = (route: Route | RouteCollection, prefixedPath: string) => (
+    'initialPath' in route ? '' : prefixedPath
+  );
```
wSedlacek, line 29: We should probably rename these functions.
`get` implies some type of network or storage transaction which isn't hat is happening here.

The `getPrefixedPath()` is only used in one place so we could just inline that or simply put it in a variable.
I am open to suggestions but `getChildPath()` and `getPrefixedPath()` need better names.

wSedlacek (commented):
These changes look good. Now that we are starting to get this to be more stable we should start updating some of the badly named functions as we clean things up.

Do you have any ideas for good names for `getChildPath()` or `getPrefixedPath()`?

packages/fsapp/src/beta-app/router/types.ts
```diff
@@ -175,7 +175,10 @@ export type Route =
  */
 export type Routes = readonly (Route | RouteCollection)[];
 
-export type ExternalRoute = Route & { readonly tabAffinity?: string };
+export type ExternalRoute = Route & {
+  readonly tabAffinity?: string;
+  readonly statusBarStyle?: OptionsStatusBar;
```
wSedlacek, line 180: Is this necessary when `readonly statusBarStyle?: OptionsStatusBar;` exists on `ComponentRoute` which is part of the `Route` union?

I recall you mentioning there was some sort of type error, but what were the conditions of that?
Maybe we can solve that with type guards?

wSedlacek (approved):
LGTM

bweissbart (approved):

PR #2061, opened 2021-05-05
fix(fsapp): pass keys from env to dev menu for SInfo

griffithcaleb:
**Description**
- The current implementation of the Storage Manager passes an empty object as the `RNSensitiveInfoOptions` in the dev menu of flagship. 
- Because of this, items that were stored using specific `RNSensitiveInfoOptions` such as 
`
const options: RNSensitiveInfoOptions = {
  sharedPreferencesName: 'xxx-Prefs',
  keychainService: 'xxx-Service',
};` 
are not deleted because the same options must be passed to RNSInfo to delete the stored items. 
- adding option to add these `RNSensitiveInfoOptions ` to the env file so we can access them in the dev menu to successfully delete items that were stored in RNSInfo.


bweissbart (changes requested):
Can you please add an optional prop to our environment typing for sInfoKeys? https://github.com/brandingbrand/flagship/blob/develop/packages/flagship/src/types.ts#L68

JeromeFitz:
Closing in favor of #2113

PR #2088, opened 2021-05-27
feat: public js docs

wSedlacek:
## Description

This makes the comments public so that JS Doc comments are not removed from bundles.
This means that hover over docs in VS Code will work.
This means that Serializable Components will not have their `TJS-Ignore` comments removed﻿


winescout:
they still get scrubbed in prod web pack though?

wSedlacek:
> they still get scrubbed in prod web pack though?

That is correct. Webpack and Metro would scrub them.

bweissbart (approved):

PR #2094, opened 2021-06-02
feat(fscomponents): FS-2308 - allow serializableImage to have children

andyschwob:
This replaces the plain SerializableImage component to one that accept children and can act as an overlay.  If you haven't used the `ImageBackground` component much before the code for it is here: https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageBackground.js

It's basically a wrapped/styled Image component and it does have web support.

- remove height/width from the image `source`.  We really should never have been passing it in with the source object, it's why we need the ImagePropsTransformer in the app to undo this.  Instead it should only be in the style object
- I wrapped `ImageBackground` in another view because it allows padding (and possibly a background color) for the image, which is I believe what a user would expect when adding padding (see picture).  Because of `StyleSheet.absoluteFill` being applied to the Image in `ImageBackground` we need to pass the `style` into `ImageBackground` as well.

I've tested it in a variety of different styles of overlay and seems to work well in creating what we need to create.

### DIFF
--- a/packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
+++ b/packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
@@ -35,12 +35,12 @@ export interface PreStandardizedSerializableImageProps
 export type SerializableImageProps = StandardContainerProps<PreStandardizedSerializableImageProps>;
 
 export const SerializableImage: FC<SerializableImageProps> =
-  ({ children, style = {}, uri, ...props }) => {
+  ({ children, containerStyle, style = {}, uri, ...props }) => {
     const source = useMemo(() => ({ uri }), [uri]);
     const { alignItems, justifyContent, ...restStyles } = style;
 
     return (
-      <View style={restStyles}>
+      <View style={[restStyles, containerStyle]}>
         <ImageBackground
           {...props}
           source={source}
--- a/packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
+++ b/packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
@@ -1,5 +1,5 @@
 import React, { FC, useMemo } from 'react';
-import { Image, ImageProps, ImageStyle } from 'react-native';
+import { ImageBackground, ImageProps, ImageStyle, View, ViewStyle } from 'react-native';
 
 export interface SerializableImageProps
   extends Pick<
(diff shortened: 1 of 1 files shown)


packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
```diff
@@ -14,16 +14,27 @@ export interface SerializableImageProps
...
     | 'testID'
   > {
   uri?: string;
-  height?: number;
-  width?: number;
-  style?: ImageStyle;
+  style?: ViewStyle;
+  imageStyle?: ImageStyle;
+  children?: React.ReactNode;
 }
```
wSedlacek, line 24: Can we standardize these props? (See SerializableView)
Mainly we are missing the `containerStyle` which is a style that directly wraps the child objects.

packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
```diff
@@ -14,16 +14,27 @@ export interface SerializableImageProps
...
+  ({ children, style, uri, ...props }) => {
+    const source = useMemo(() => ({ uri }), [uri]);
+
+    return (
+      <View style={style}>
+        <ImageBackground
+          {...props}
+          source={source}
+          style={[{ flex: 1 }, style]}
+        >
```
wSedlacek, line 35: It looks like we are using the `style` prop on two different components.
How this this interact when doing margins or padding?
Do they get doubled unexpectedly?

wSedlacek (changes requested):
@andyschwob can you take a look at my feedback?
I would like to get this merged in my our release later today.

packages/fscomponents/src/components/serializable/vNext/SerializableImage.tsx
wSedlacek: https://github.com/andyschwob/flagship/blob/43e16a31e845522910dbad0001ace465408fc7c6/packages/fscomponents/src/models/StandardizedComponent.ts#L15-L19

andyschwob:
@wSedlacek 

Upon further review I was not using the styles correctly.  So in order to get the image that I posted above to work correctly, the outer <View> needs to have all the components of `style` applied to it expect for the alignment of the children (alignItems & justifyContent).  Those alignment props needs to be applied to the inner View inside <ImageBackground>.  So I'm going to destructure style and apply the appropriate part of style to each element.

wSedlacek (approved):
LGTM. Great work!

PR #2104, opened 2021-06-15
feat(fscomponents): add ability to hide overflow in multi carousel

varzaman:
- added new prop to hide horizontal overflow in the carousel
- added storybook example

Accidentally pushed this up to the repo, please delete during merge.

This is specifically for flagship 10. I will make a PR for flagship 11.


wSedlacek:
> overflowX: this.props.hideOverflow ? 'hidden' : 'scroll'

What are the differences in `hidden`/`scroll` here?
Is this to remove the scrollbar?

varzaman:
Yes, setting it to hidden removes the scrollbar, which also removes the "free scrolling", and allows it to be moved left and right by the existing swipe gestures.

bweissbart (approved):

PR #2107, opened 2021-06-16
feat(fsapp): FS-2511 expose hydrate RNW config option

nathan-sankbeil:


bweissbart:
I spoke with the team this is intended for and they're migrating to FS 10 soon so we don't need this.

PR #2110, opened 2021-06-16
fix(fsapp): broken links

wSedlacek:
The following route was being registered as `shop/shop/category/:categoryId`.
```ts
{
    initialPath: 'shop',
    tab: Shop,
    children: [
      {
        path: 'shop/category/:categoryId',
        loadComponent: () => import('./screens/Category').then(({ Category }) => Category)
      },
   ]
}
```
With the `initialPath` prepended when it shouldn't have been.
This fixes that.


andyschwob (approved):
Tested on web and native, looks good to me

PR #2118, opened 2021-06-23
feat(fsapp): expose history type

wSedlacek:
## Description

This exposes the type returned by `useNavigator()` so that the navigator can be more portable for example passsing it into the modal context.


dtwiers (approved):
LGTM

PR #2173, opened 2021-07-19
feat(fscomponents): FS-2640 - Builder | Allowing setting number of items to show per breakpoint

wSedlacek:
## Description

This updates the `MultiCarousel` component to be a functional component
This updates the scrolling behavior of the `MultiCarousel` component so it is more natural on all platforms (especially web)
This adds a new porp `itemWidth` which will allow setting a specific width for items in the carousel to be.

### DIFF
--- a/packages/fscomponents/src/components/Grid.tsx
+++ b/packages/fscomponents/src/components/Grid.tsx
@@ -27,9 +27,12 @@ import { chunk } from 'lodash-es';
 const DEFAULT_COLUMNS = 2;
 const DEFAULT_MIN_COLUMNS = 175;
 const DEFAULT_BACK_TOP_BUTTON_SHOW_AT_HEIGHT = 100;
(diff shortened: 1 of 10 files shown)


bweissbart (changes requested):
Pagination seems to be an issue now. In PirateShip, our implementation of the carousel on the Shop screen has itemsPerPage [set to 1](https://github.com/brandingbrand/flagship/blob/develop/packages/pirateship/src/components/PSHeroCarousel.tsx#L91) but the carousel is now showing all items:

image

The expected behavior in this instance is that only one of the images would display and would fill the entire container.

wSedlacek:
> carousel on the Shop screen has itemsPerPage set to 1 but the carousel is now showing all items
Good catch, let me see if I can identify what happened.

wSedlacek:
@bweissbart  I have corrected the issue.
Screen Shot 2021-07-20 at 8 16 01 AM

bweissbart (approved):

PR #2183, opened 2021-08-02
feat: static context reference

wSedlacek:
## Description

This adds in a dependency injection library `fslinker`
This updates all context within `fsapp` to be statically referenced by the `fslinker` library
This assures that references will be maintained even when used across bundles
This adds in several `InjectionTokens` for things like the redux store, the API defined for the app, and the navigator, so that they can be injected anywhere regardless of context.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ packages/pirateship/android
 packages/pirateship/ios
 packages/pirateship/web
 packages/pirateship/env/env.js
+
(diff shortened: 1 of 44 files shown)


dtwiers (approved):
Given the size of this PR, the following seems inadequate, but I'll do it anyway:

LGTM

nathan-sankbeil (approved):

PR #2195, opened 2021-08-12
fix(fsapp): fix modal options typing

buticm:
## Description

- makeModal uses optional parameter for component's options.
It has to include options from react-native-navigation package that are allowed for modals.

## Test
- use together with project that makes use of `makeModal`
- try to use option `modalPresentationStyle: OptionsModalPresentationStyle.fullScreen`
- run `yarn tsc`

ER: no errors


packages/fsapp/src/beta-app/modal/types.ts
```diff
@@ -10,7 +10,7 @@ export interface ModalService {
   dismissAllModals(): Promise<void>;
 }
 
-export interface ModalOptions extends RNModalOptions, ModalProps {
+export interface ModalOptions extends RNModalOptions, ModalProps, Options {
```
wSedlacek, line 13: Adding `Options` to this type will not make the options automatically apply.

This change would require changes to the modal.context.tsx so that it properly applies the options set here.
https://github.com/brandingbrand/flagship/blob/69e99e0e7841f116fdb839833783633b7ea38ce7/packages/fsapp/src/beta-app/modal/modal.context.tsx#L89-L92

wSedlacek (changes requested):
@buticm Can you make the changes to `modal.context.tsx` so that the options are applied?

packages/fsapp/src/beta-app/modal/types.ts
buticm: @wSedlacek,
I didn't add any new options.
I only fixed the typing.
Whatever passed in the `modal.options` now may include properties defined in `Options` from `react-native-navigation`.
The way it's passed to `Navigation.showModal` doesn't need to change.

wSedlacek: The `modal.options` get's passed to this `.showModal()` call.
```tsx
      Navigation.showModal({
          stack: {
            id: MODALS_STACK,
            children: [
              {
                component: {
                  id,
                  name: modal.definitionId,
                  options: {
                    modal: modal.options,
                    topBar: modal.topBarOptions
                  },
                  passProps: {
                    resolve,
                    reject,
                    ...props
                  }
                }
              }
            ]
          }
        });
      });
```

You will notice that `modal.options` is assigned to the `modal` property in the `options` record.
This property has the following type. [Source](https://wix.github.io/react-native-navigation/6.12.2/api/options-root#modal)

```ts
export interface ModalOptions {
    /**
     * Control wether this modal should be dismiss using swipe gesture when the modalPresentationStyle = 'pageSheet'
     * #### (iOS specific)
     */
    swipeToDismiss?: boolean;
}
```

So for something like `modalPresentationStyle` to be applied you would need something like
```tsx
      Navigation.showModal({
          stack: {
            id: MODALS_STACK,
            children: [
              {
                component: {
                  id,
                  name: modal.definitionId,
                  options: {
                    modal: modal.options,
                    topBar: modal.topBarOptions,
                    modalPresentationStyle: modal.options.modalPresentationStyle
                  },
                  passProps: {
                    resolve,
                    reject,
                    ...props
                  }
                }
              }
            ]
          }
        });
      });
```

buticm:
@wSedlacek,
I see the issue now.
Somehow `Navigation.showModal` actually allowed options to be set inside `options.modal` for some reason despite how the types are defined.

I updated the typing and application of the options.

packages/fsapp/src/beta-app/modal/types.ts
```diff
@@ -13,6 +13,7 @@ export interface ModalService {
 export interface ModalOptions extends RNModalOptions, ModalProps {
   readonly backdropStyle?: StyleProp<ViewStyle>;
   readonly title?: string;
+  navigationOptions?: Options;
```
wSedlacek, line 16: This should probably be `readonly` like the other options.
Not a big deal, just a preference to assure values are not unexpectedly changed.

packages/fsapp/src/beta-app/modal/modal.context.tsx
```diff
@@ -87,8 +87,15 @@ export const ModalProvider: FC<ModalProviderProps> = ({ children, screenWrap })
...
-                    topBar: modal.topBarOptions
+                    ...modal.options?.navigationOptions,
+                    modal: {
+                      ...modal.options?.navigationOptions?.modal,
+                      ...modal.options
+                    },
+                    topBar: {
+                      ...modal.options?.navigationOptions?.topBar,
+                      ...modal.topBarOptions
+                    }
```
wSedlacek, line 98: It feels a bit weird to include so many extra values in `options.modal`, ie `navigationOptions`.
This will behave as you expect it to, there will just be extra values in there that are unused.
That said we were already doing that with the web props so probably not a big deal.

packages/fsapp/src/beta-app/modal/types.ts
wSedlacek: It feels a bit strange to be able to set the `RNModalOptions` both directly on `ModalOptions` and in `navigationOptions`.
Granted removing it from `ModalOptions` would be a breaking change, but this will probably be refactored before fs12 to clean up the interface.

wSedlacek (approved):
> Somehow Navigation.showModal actually allowed options to be set inside options.modal
TypeScript will do that. It uses Duck typing, so if you add extra keys it won't complain so long as the keys it needs are there.
We might get a feature if future versions of TypeScript to catch things like this.

This code should work, the interface is a bit confusing, but not a big deal for the fs11 release.

bweissbart:
@buticm - can you please make a corresponding PR or cherry pick this into master so I can get you a v10 release?

wSedlacek:
These changes are for beta-app so wouldn't be relevant to v10

PR #2199, opened 2021-08-17
fix(fsapp): web location initialization

wSedlacek:
## Description

This fixes the activation initial activation of routes on web.
Many screens are designed such that they assume that all values will be rsolved before mounting.
Previously this was not the case as the location would be set by the browser's location loading routes before the resolves and activation was complete.

This fixes that problem by providing a place holder location of `__SPLASH__` to be used until activaiton is complete.
This `__SPLASH__` route can also be registered through the router to controll what is shown during this activation time.﻿


packages/fsapp/src/beta-app/router/history/history.web.ts
```diff
@@ -50,7 +50,13 @@ export class History implements FSRouterHistory {
 
   private _action: Action = this.browserHistory.action;
 
-  private _location: Location = Object.freeze(cloneDeep(this.browserHistory.location));
+  private _location: Location = Object.freeze({
```
bweissbart, line 53: I was looking up the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) for Object.freeze and saw this:

> freeze() returns the same object that was passed into the function. It does not create a frozen copy.

As such, it would seem to me that this is an unnecessary call.

wSedlacek: The intention here wasn't to create a copy, but to make the object readonly.
The `_location` is intended to be replaced as a whole and never have individual values change,
ie `this._location.path = "something"`
Changing a value like that would not update the reference to `_location` and thus consumers using like in `useEffect()` would not rerun.

While the code written in this class does not update `_location` like that, this was put simply to avoid an update causing a regression in the future. Given the complexity of the beta-app I thought it prudent to freeze the object.

I would prefer to keep the `Object.freeze()` here, but if you insist I will remove it.

bweissbart:
@wSedlacek - I'm okay keeping your code as-is but unfortunately there are conflicts now.

wSedlacek:
@bweissbart Resolved the conflicts

PR #2209, opened 2021-08-27
chore: use exact version dependencies

wSedlacek:
## Description

Because we have not been using semver for our alpha releases we commonly have releases that have breaking changes that are not captured by ^ versioning. Even ~ versioning would not capture these.

To combat downstream projects getting uexpected versions, this sets all dependnecies to use exact versions. So if you get alpha.15 of fsapp then that will have alpha.15 of fslinker


bweissbart:
Per our research after this was made, Lerna publish controls how this is managed. When we do a alpha release in the future we'll have to remember to append the `--exact` flag.

PR #2214, opened 2021-09-01
fix(fscomponents): rerender when render changes

wSedlacek:
## Description

This updates the Grid component to rerender when it's render has changed

### DIFF
--- a/packages/fscomponents/src/components/Grid.tsx
+++ b/packages/fscomponents/src/components/Grid.tsx
@@ -459,6 +459,10 @@ export const Grid = <ItemT, >(props: GridProps<ItemT>) => {
     ]
   );
 
+  const rerenderTrigger = useMemo(() => {
+    return Symbol('This is used to mark a change that requires a rerender');
+  }, [listViewProps?.extraData, renderRow]);
+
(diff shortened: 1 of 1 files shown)


packages/fscomponents/src/components/Grid.tsx
```diff
@@ -484,6 +484,7 @@ export const Grid = <ItemT, >(props: GridProps<ItemT>) => {
         onEndReachedThreshold={onEndReachedThreshold}
         onLayout={onLayout}
         onContentSizeChange={setWidth}
+        extraData={[...listViewProps?.extraData, renderRow]}
```
bweissbart, line 487: If extraData is undefined this is going to error when we try to spread on it. Should we add a check for that?

bweissbart (changes requested):

packages/fscomponents/src/components/Grid.tsx
wSedlacek: Oh! extraData is `any` so I wasn't getting errors for that.
Let me fix this.

bweissbart (approved):

PR #2225, opened 2021-09-14
feat(fsapp): get initial url

wSedlacek:
## Description

This allows apps to use custom logic for deep link initialization

### DIFF
--- a/packages/fsapp/src/beta-app/app/app.ts
+++ b/packages/fsapp/src/beta-app/app/app.ts
@@ -14,7 +14,10 @@ export class FSAppBeta extends FSAppBase {
   public async startApplication(): Promise<void> {
     Linking.addEventListener('url', ({ url }) => this.router.open(url));
 
-    const url = await Linking.getInitialURL();
+    const url = await(
+      this.config.getInitialURL ? this.config.getInitialURL() : Linking.getInitialURL()
+    );
+
     if (url) {
       await this.router.open(url);
     } else {
--- a/packages/fsapp/src/beta-app/app/types.ts
+++ b/packages/fsapp/src/beta-app/app/types.ts
@@ -33,6 +33,7 @@ export interface AppConfig<
       ? StoreConfig<S, A>
       : undefined
     : undefined;
+  readonly getInitialURL?: () => Promise<string | null>;
 }
 
 export interface IApp {


crherman7 (approved):

packages/fsapp/src/beta-app/app/app.ts
```diff
@@ -14,7 +14,10 @@ export class FSAppBeta extends FSAppBase {
   public async startApplication(): Promise<void> {
     Linking.addEventListener('url', ({ url }) => this.router.open(url));
 
-    const url = await Linking.getInitialURL();
+    const url = await(
+      this.config.getInitialURL ? this.config.getInitialURL() : Linking.getInitialURL()
```
bweissbart, line 18: Will getInitialURL always exist on config or do we need to do an optional chaining?

bweissbart (changes requested):

packages/fsapp/src/beta-app/app/app.ts
crherman7: `this.config` would exist correct? So if you are checking the value here `this.config.getInitialURL` would be safe because at a minimum it would just be `undefined` - you aren't accessing anything on the attribute until the ternary clause. Ideally this should be fine - but optional chaining is always safe.

wSedlacek: Chris is correct, `this.config` is not nullable, so you can access any property on the first level on it without reference errors and the returned result would be undefined.
TypeScript checks for this in `strict` mode so long as you aren't loosing type safety anywhere.

Where we have turn into trouble in the past is with using indexes like `this.config.routes[0].path`
Indexes can always return `undefined` but TypeScript does not check for this unless you turn on [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) which was added in 4.1 iirc, but we have not enabled it yet.

bweissbart: Correct, config should always exist, but if getInitialURL doesn't exist it'll throw an error because you're trying to call it as a function. Instead you could do `this.config.getInitialURL?.()`

wSedlacek: So the condition of the ternary, `this.config.getInitialURL` would check if it is truthy, and thus in the first part of the ternary the value must be truthy.

Given that the only valid types of `this.config.getInitialURL` are `() => Promise<string | null>` or `undefined` if it is truthy it must be a function that returns a promise, unless someone used `any`.

Furthermore, the ternary is used here rather than optional chaining + null coalescing to give the option to opt out of the default behavior, example.
```ts
this.config.getInitialURL?.() ?? Linking.getInitialURL()
``` 
If `getInitialURL` returned `null` then `Linking.getInitialURL()` would be called.
This would mean that `getInitialURL` must either return a path OR simply fallback.

Where as with this
```ts
this.config.getInitialURL ? this.config.getInitialURL() : Linking.getInitialURL()
```
If `getInitialURL` exists then it will be called and can return `null` without `Linking.getInitialURL()` getting invoked as a fallback.
This makes the `getInitialURL()` completely in charge of getting the initialURL making whatever it returns explicitly what will be opened and give more expected behavior for those using the API.

wSedlacek: Here is a minimal recreation in [TypeScript Playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM7IDeAUMmcntgLYQBiAriAlrgPwBcyAFAJTIC8APgpgooAObIAPshD0ANvIDcxAL7FiCXHjDJsmHPgCMnDC3wCiqlVvy795vACZTB7ZZLkK1Oo2aHOXgFhAHJKGjAACwkQtRVNbXs3fABmV0cPUnJw3yZzQL4hWQV5OI1bHWQoRjNDPAAVaBA4KABPSy4HOvS6wuFPcigIMHooEG4u7QA6HIY8w2RWPWS8GZ85-1wg9hD4RQAjRABrEJ4VdQS7KpqV2tw4eQBhSLgJCQ7J-B7tPqIssiGIzGy0caxoG3MrCmQVYS12D3khwQJ3O5W02HkECm8mw4i4ACJHnBxoDRuM5Ip8WdLpRMdjcVxqiA7vhGmMWq1OisjDxqRUMVicXimSyGk0OVzHE5eTZ0XShYybo42c02pK6ikZWj8AL6Xj8bRsFAkAATZBgbDIPaI45U2U6+UMkW3ZIPZ6vEDiCTq7Q8vlywVOpV1Flul5vT0+-DS-0OwPC4PaUNPcOe72fPCas5AA) to play with.

bweissbart: Many apologies -- I completely overlooked that the entire ternary was based on the existence of getInitialURL. I must've been really tired. This makes 100% sense looking at it with fresh eyes.

bweissbart (approved):

packages/fsapp/src/beta-app/app/app.ts
wSedlacek: No worries! Thank you for double checking my logic.

PR #2242, opened 2021-10-12
fix(fsengagement): fix RenderImageTextItem deeplinks

laine-hallot:
Im just going to copy this from are related PR I made on another project since it describes what I did pretty well.

> ## Description
> 
> `RenderImageTextItem` couldn't access `EngagementContext` which meant that `handleAction` was undefined and calling it when the user tapped the image would fail. `RenderImageTextItem` should have had access to the context since its supposed to be a child of `EngagementComp` but I couldn't find as a descendant anywhere in the code. To get around that I went ahead and converted `RenderImageTextItem` from a class component to a functional component so that I could give it access to the right context using the `useContext()` hook. Everything seems to be working after the conversion but if `RenderImageTextItem` was a class component for a reason and switching it causes problems that I'm not aware of please let me know so I can figure out a different fix.

### DIFF
--- a/packages/fsengagement/src/carousel/RenderImageTextItem.tsx
+++ b/packages/fsengagement/src/carousel/RenderImageTextItem.tsx
@@ -1,10 +1,8 @@
 /* tslint:disable */
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
+import React, { useContext, useState  } from 'react';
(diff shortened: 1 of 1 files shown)


wSedlacek:
Already fixed by @andyschwob in prior work.

PR #2245, opened 2021-10-18
feat(fscomponents): adds function prop to TabbedContainer

coronary:
adds function prop onTabSwitch to TabbedContainer


skyeckstrom (approved):

wSedlacek (approved):
The changes here look good. Thank you for your contribution @coronary

PR #2263, opened 2021-11-09
feat: update web carousel dependencies

bweissbart:
This updates the Swiper dependency to fix a security issue. The fixed version is a breaking change from the previous (v5 to v6) which in turn broke our custom implementation which uses react-id-swiper in conjunction with Swiper. (The pagination buttons were no longer displaying.)

I refactored the web carousel code to use the "custom build" solution described in the react-id-swiper readme and it solved the problem: https://www.npmjs.com/package/react-id-swiper#custom-build-swiper

This also adds a new screen to the dev menu to test carousels.

There are two ways to test this:

1. Run `yarn dev:storybook` and play with the Carousel stories
2. Run PirateShip web (`yarn ship:compile-web`), go to the develop menu, and then go to the Carousel sample screen

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "marked": "^0.7.0",
     "websocket-extensions": "^0.1.4",
     "react-dev-utils": "^11.0.0",
-    "immer": "^8.0.1"
+    "immer": "^9.0.6"
   },
   "workspaces": [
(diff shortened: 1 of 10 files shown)


nathan-sankbeil (approved):
lgtm

PR #2268, opened 2021-11-10
fix(flagship): stop forcing lowercasing when renaming packages

### LINKED ISSUE #1778: Cannot init on case sensitive file systems
https://github.com/brandingbrand/flagship/blob/cad5d7ee97d22879fe7c3633d4e405783382dc2b/packages/flagship/src/lib/rename.ts#L55

When attempting to run the project on case sensitive filesystems this `.toLowerCase()` results in the file path to be incorrect if any of the parts of the path are upper case. In testing I have found that simply removing the `.toLowerCase()` resolves this behavior.

skyeckstrom:
We were unnecessarily forcing the path to lowercase when renaming packages. This would break on case sensitive filesystems if you had a capital letter in the path containing your project.

Fixes #1778

### DIFF
--- a/packages/flagship/src/lib/rename.ts
+++ b/packages/flagship/src/lib/rename.ts
@@ -52,7 +52,7 @@ export function pkgDirectory(oldPkg: string, newPkg: string, ...pathComponents: 
 
     // Rename matching paths
     results.forEach(oldPath => {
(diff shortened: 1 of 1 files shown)


wSedlacek (approved):
LGTM

PR #2270, opened 2021-11-10
chore: switch to react-native-async-storage version

skyeckstrom:
async-storage has moved from react-native-community to react-native-async-storage.


bweissbart:
Need to do some validation on this one. Because this has a native dependency we could get into issues if projects have different versions of this dependency. Also not sure how the scoping would affect this.

bweissbart:
Because of breaking change concerns with this PR, I merged it into my WIP Flagship 12 branch: https://github.com/bweissbart/flagship/commit/73fb2647c39619b3ae26f29a5a5257502bb254ff

PR #2272, opened 2021-11-10
fix(fslinker): update global typing to support node 16

skyeckstrom:
The DefinitelyTyped definition of the global object for Node 16 removed NodeJS.Global as an exported type for the global object, which causes this file to fail to compile.

This change explicitly declares `__FLAGSHIP_LINKER_GLOBAL_CACHE__` as existing in the global scope so we can compile for Node 14 and 16.


wSedlacek (approved):
LGTM

PR #2278, opened 2021-11-10
chore: remove unused push-notification-ios

skyeckstrom:
Removes unused @react-native-community/push-notification-ios package.

Closes #2217


bweissbart (dismissed):

sjotterman (approved):
Looks good to me, but I'm not an iOS expert - when would this AppDelegate be run, and how do I test it?

PR #2319, opened 2021-11-17
feat(fscomponents): Add Autoplay to Carousel

ckornell:
* Adds autoplay with a Knight Rider effect
* Allows to customize duration
* Stops on interact, including tapping or clicking the slide
* Disables animations for autoplay transition durations of less than 300ms


packages/fscomponents/src/components/MultiCarousel/MultiCarousel.web.tsx
```diff
@@ -286,6 +296,43 @@ export const MultiCarousel = <ItemT, >(props: MultiCarouselProps<ItemT>) => {
     await goTo(currentIndex, options);
   }, [goTo, currentIndex]);
 
+  const goToNextCancelCarousel = async () => {
```
ckornell, line 299: `goToNextCancelLoop` would be more appropriate.

wSedlacek:
@ckornell Can we get a video clip of this change?

packages/fscomponents/src/components/MultiCarousel/MultiCarousel.web.tsx
```diff
@@ -107,6 +107,8 @@ export const MultiCarousel = <ItemT, >(props: MultiCarouselProps<ItemT>) => {
     hideScrollbar,
     itemStyle,
     itemsAreEqual,
+    autoplay = true,
```
bweissbart, line 110: Did the previous version of this carousel have some form of autoplay by default? If not then we should default this to false.

bweissbart (changes requested):

packages/fscomponents/src/components/MultiCarousel/MultiCarousel.web.tsx
wSedlacek: Having this not as the default will require additional submarine changes so that the value can be set.
Do we want to do that instead?

ckornell:
> @ckornell Can we get a video clip of this change?

I'm looking for the CMS slot these images our powered by. Sorry for the bad demo.

https://user-images.githubusercontent.com/2124870/142277721-9fbb21d3-2327-4286-9e73-83d8296567b9.mp4

wSedlacek (dismissed):

packages/fscomponents/src/components/MultiCarousel/MultiCarousel.web.tsx
```diff
@@ -286,6 +296,43 @@ export const MultiCarousel = <ItemT, >(props: MultiCarouselProps<ItemT>) => {
...
+
+  useEffect(() => {
+    if (autoplay && numberOfPages > 0 && !animating && shouldPlay) {
+      autoplayTimeout.current = setTimeout(async () => {
+        const nextIndex = currentIndex + 1;
+        const options = {
+          animated: true
+        };
+        // Animation is hardcoded at 200ms; stop animation to avoid glitch.
+        if (autoplayTimeoutDuration >= 300) {
```
ckornell, line 317: this is incorrect pushing a fix now

wSedlacek (approved):

bweissbart (approved):

PR #2324, opened 2021-11-19
Autoscroll animations

winescout:
In this PR
==========
Updates autoscrolling to snap back to first image when it gets to the end
Adds a autoscrolling carousel to the playbook

https://www.screencast.com/t/yOc9hCvg


ckornell:
A few things: 

1. I disagree that going back to the first slide without animation is a good idea. In my opinion, It's actually very jolting and does not seem intentional.

2. Because you modified the `goToNext` function, and always check for a zero index to turn off animation, it breaks the arrow functionality and never animates `goToNext` from the first slide; it also confuses what that function is supposed to be responsible for, IMO.

3. I've reopened my PR, and I think that should be considered as the solution here.

Here's some videos to show this PR. The first one shows the jump, the second one shows the broken arrow functionality.

https://user-images.githubusercontent.com/2124870/142624922-be4bfb4e-21e3-4f3d-8720-25b26564e4c1.mp4


https://user-images.githubusercontent.com/2124870/142624929-bdde3639-d52e-4826-b9ef-407e8e72cc55.mp4

packages/fscomponents/src/components/MultiCarousel/MultiCarousel.web.tsx
```diff
@@ -281,9 +281,13 @@ export const MultiCarousel = <ItemT, >(props: MultiCarouselProps<ItemT>) => {
   );
 
   const goToNext = useCallback(async (options?: GoToOptions | GestureResponderEvent) => {
-    const nextIndex = currentIndex + 1 > numberOfPages - 1 ? numberOfPages - 1 : currentIndex + 1;
+    const nextIndex = currentIndex + 1 > numberOfPages - 1 ? 0 : currentIndex + 1;
+    const animationOptions =
+      nextIndex !== 0 &&
```
ckornell, line 286: Other functions in this component use this function to advance the pages. Here you are always turning off animation if the nextIndex is 0. This means even without autoplay, if a user wants to use animation for the arrows, it will never work. It appears very broken. You would need to remove this logic and handle this in the useEffect or create another function that the useEffect calls. In any case, from a UX perspective...having transitions and those transitions suddenly drop out does not seem intentional and can be very jolting to some users.

winescout:
I need to re-open this PR with cleaned up commit messages.  @ckornell Let's pair up this morning and find a solution together.

winescout:
New PR:  https://github.com/brandingbrand/flagship/pull/2325

PR #2327, opened 2021-11-22
fix: exported currency value from it's own type file instead of index

chrisjlan89:
Tciket:
https://brandingbrand.atlassian.net/browse/FS-3207

### DIFF
--- a/packages/fscommerce/src/Commerce/CommerceTypes.ts
+++ b/packages/fscommerce/src/Commerce/CommerceTypes.ts
@@ -6,7 +6,7 @@ export * from './types/Cart';
 export * from './types/CartQuery';
 export * from './types/Category';
 export * from './types/CategoryQuery';
-export * from './types/CurrencyValue'
+export * from './types/CurrencyValue';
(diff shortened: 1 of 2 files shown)


bweissbart (approved):

packages/fscommerce/src/Commerce/CommerceTypes.ts
```diff
@@ -11,6 +6,7 @@ export * from './types/Cart';
 export * from './types/CartQuery';
 export * from './types/Category';
 export * from './types/CategoryQuery';
+export * from './types/CurrencyValue'
```
bweissbart, line 9: Need a semicolon here.

bweissbart (changes requested):

packages/fscommerce/src/Commerce/CommerceTypes.ts
wSedlacek:
```suggestion
export * from './types/CurrencyValue';
```

bweissbart (approved):

PR #2399, opened 2022-08-02
feat: AS-297 Make refine header optional

HosMohamed:
This PR adds a new props to the `FilterListDrilldown` to allow for hiding the refine bar. 

new props defaults to `true` and it's named `showDrillDownHeader` (Optional- boolean).


crherman7:
This package will no longer be a part of this repo. This will be handled in a separate repo where further fscomponents refactor will be completed.

PR #2400, opened 2022-08-03
feat(VBMS): SEC-1698 Add dependabot config 🤖

xero:
https://brandingbrand.atlassian.net/browse/SEC-1698


bweissbart:
We don't manage any code in this repo directly anymore -- we use shipit to port code from shipyard onto this repo.

PR #2444, opened 2022-12-22
FCR-3668 plugin-fastlane

danrudolf2:
## Ticket
[FCR-3668](https://brandingbrand.atlassian.net/browse/FCR-3668)

## Description
 - Adding plugin-fastlane

[FCR-3668]: https://brandingbrand.atlassian.net/browse/FCR-3668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/react-native/.kernelrc/env/env.prod.ts
```diff
@@ -1,15 +1,30 @@
...
+      provisioningProfileName: "Test Provisioning Profile",
+      profilesDir: "./xx"
+    }
   },
   android: {
     displayName: 'Hello World',
     packageName: 'com.helloworld',
+    versioning: {
+      version: '0.0.1',
+      build: 1,
```
NickBurkhartBB, line 26: nothing seems to be relying on these build numbers, the fastlane file just seems to rely on fetching from app center, and this doesn't seem to get injected into android build.gradle
Should we drop it from the type? Or use it if no version yet in app center?

NickBurkhartBB:
this plugin seems appcenter specific, should it be called plugin-fastlane-appcenter?

apps/react-native/.kernelrc/env/env.prod.ts
danrudolf2: yeah i should probably change the name of the android one to `versionCode`. its not necessarily relevant to app-center which is why I didn't do the version in the `kernelPluginFastlane` both version and versionCode should end up in the `app/build.gradle`

crherman7: We will need to discuss this because if there is a present build number it will override the versionCode that is fetched from app center i.e. disable that part of fastfile

crherman7 (commented):
Things that need to be done:
- unit tests
- update build command to copy over template to dist folder i.e. postbuild - otherwise it won't get bundled

packages/core/src/utils/path.ts
```diff
@@ -66,6 +66,14 @@ const resolvePathFromProject = (...paths: string[]): string =>
  */
 const getPackagePath = () => resolvePathFromProject("package.json");
 
+/**
+ * Resolves a path to the package.json
+ *
+ * @returns A path to the package.json relative to project path
+ */
+const getPathToPlugin = (pluginName: string) =>
+  path.resolve.apply(__dirname, ["..", "..", "packages", pluginName]);
```
crherman7, line 75: this only seems relative to the monorepo not an actual react native project.....was this to fulfill the unit test? if so we need to use mocks instead of creating code based around unit test

packages/plugin-fastlane/package.json
```diff
@@ -22,6 +22,7 @@
   ],
   "scripts": {
     "build": "microbundle --target node --compress --format cjs,esm --tsconfig tsconfig.build.json --no-sourcemap",
+    "postbuild": "cp -r src/fastlane.template dist/fastlane.template",
```
crherman7, line 25: `dist/fastlane.template` -> `dist/fastlane`

packages/plugin-fastlane/src/index.ts
```diff
@@ -1,9 +1,73 @@
...
+  pathConfig: PathConfig,
+  config: Config & KernelPluginFastlane
+) => {
+  const templateDir = path.resolve(
+    path.project.pluginPath("plugin-fastlane"),
+    "src",
+    "fastlane.template",
+    pathConfig.platform,
+    "fastlane"
+  );
```
crherman7, line 22: this does not seem correct.....remember this is going to be bundled i.e. a single js file inside the dist folder - the template folder will be in the same directory

packages/plugin-fastlane/src/index.ts
```diff
@@ -1,9 +1,73 @@
...
+
+  const targetDir = path.resolve(pathConfig.path.rootDirPath(), "fastlane");
+  if (!fs.existsSync(targetDir)) {
+    fs.mkdirSync(targetDir);
+  }
```
crherman7, line 27: NIT: no need to change but we can assume this folder does not exist

crherman7 (commented):
A few additional things besides comments:
- template files do not need to have a suffix `*.template`
- you don't need to have single copy file lines you could just do a traversal similar what was done with the native template

crherman7 (approved):

PR #2465, opened 2023-01-06
Initial FS12 Env Plugin Package

seanirby:
On init, this plugin substitutes the `initialEnvName` value in the native bridges with the one provided via command line arguments in the command line arguments.  If one is not supplied the default is 'dev'.


crherman7:
This will be propagated to a new repo soon.

PR #2476, opened 2023-01-24
feat(fs12): add docs

crherman7:


jasonmosley:
This looks like a really good foundation for our documentation work. Let me know when this is ready to be reviewed.

jasonmosley (changes requested):
Could you move the content to a separate pr? I see some changes we should make but I don't want to hold up getting the doc foundation merged.

PR #2520, opened 2023-02-09
feat(fs12): android xml styles parser

crherman7:


jasonmosley (approved):
This looks good to me and my only call out is this will need to be documented in the FS12 migration docs since we are switching the default theme to day/night to match iOS. So users will have to set the style config for android.

PR #2521, opened 2023-02-10
feat(fs12): android xml strings parser

crherman7:


jasonmosley (approved):
I tested this and worked as expected. I think this is a good thing to expose but will need to be clearly documented with maybe an example use case. I am liking the direction of being able to update these files in a standard way.

PR #2524, opened 2023-02-14
fix(fs12): slim executors

crherman7:
The purpose for this change is to reduce the amount of bloat related to executors.

Previously we had pre, prePlatform, platform, postPlatform, post - the only reason for the prePlatform and postPlatform was to suffice requirements of having the ios and android directories be available before running plugins. This 100% related to running the react-native-navigation script - they have a shared path file that is required for both platforms and path functions are immediately ran. This means that the next time a platform ran without this executor format the next platform would not have those files available which would fail the script.

To remove the bloated executors but also fulfill this requirement for react-native-navigation there are 2 options. The first would be to delete the require cache related to the react-native-navigation path module but this increased build time by ~10s which is unacceptable. The second and proposed solution is to reorder the async / await for loop to run the platforms for each plugin first instead of running the plugins for each plugin.


jasonmosley (approved):
This is working as expected and RNN pluggin is working without issue with this change.

PR #2526, opened 2023-02-14
feat(fs12): info plist parser

crherman7:
Built on https://github.com/brandingbrand/flagship/pull/2525.

We want to encourage parsing parceable files over utilizing ejs for complex transformations. Ejs is good to use for simple plug-n-chug but not for complex assignments. This change adds functionality to parse plist files.


jasonmosley (approved):
This looks good and works as expected. Follow-up PRs would be unit tests and updating other plugins to use this pattern instead of ejs.

PR #2528, opened 2023-02-15
feat(fs12): execution summary table

crherman7:


jasonmosley (approved):
I like this and the changes you made after our conversation. Having a cleaner table will help make the errors stand out when they happen. 

I was trying to think of a way to have the plugin name and platform auto-discovered so you wouldn't have to add it to every plugin. But things like `Function.prototype.name` are not appropriate in production code.

crherman7:
> I like this and the changes you made after our conversation. Having a cleaner table will help make the errors stand out when they happen.
> 
> I was trying to think of a way to have the plugin name and platform auto-discovered so you wouldn't have to add it to every plugin. But things like `Function.prototype.name` are not appropriate in production code.

It's not based off the function name i.e. every function name is a named export `execute` so that is can be autrun bound by interface by cli. The filename is not appropriate because the code is bundled i.e. everything is bundled into a single index.js file.

jasonmosley:
Yeah there's just no clever way to have the log be auto aware. So going forward we will need to make sure developers follow the same naming conventions when using with Summary.

PR #2529, opened 2023-02-15
chore(fs12): remove component-dev package

crherman7:


jasonmosley (approved):
This is being removed since it will be part of phase 2.

PR #2536, opened 2023-02-22
fix(fs12): plugin parsers

crherman7:


jasonmosley (approved):
This is working as expected. This new approach is a lot cleaner and safer.

packages/plugin-leanplum/src/index.ts
```diff
@@ -66,7 +91,7 @@ const android = summary.withSummary(
 
     await fsk.update(
       path.android.mainApplicationPath(config),
-      /(^package [\s\S]+?;)/,
+      /(^package [\s\S]?;)/,
```
crherman7: undo this

PR #2540, opened 2023-02-26
fix(fs12): react-native-* nohoist policy

crherman7:


jasonmosley (approved):
This looks good to me.

PR #2541, opened 2023-02-28
chore(fs12): decompress bundles

crherman7:


jasonmosley (approved):
Given that this is a dev dependency and would not be bundled with the app compression is not really needed. I like that idea of making is easier for developers to make patches if they are needed. This is good to merge.

PR #2547, opened 2023-03-06
chore(fs12): @leanplum/react-native-sdk v2.0.0

crherman7:


jasonmosley (approved):
Perfect. Thanks

jasonmosley (commented):
Should we update the docs to let users know what version of LP we are targeting? Or should we always try to target the latest? It might be helpful for people to roll pack to an older version of a plugin.

crherman7:
> Should we update the docs to let users know what version of LP we are targeting? Or should we always try to target the latest? It might be helpful for people to roll pack to an older version of a plugin.

Correct, we need to generate new docs for the next release after the re-design.

PR #2548, opened 2023-03-06
chore(fs12): docs re-design

crherman7:


jasonmosley (approved):
Looks good so far. There are still design updates that need to be made but this completes about 90% of the needed changes. 

Since we need to have an updated doc site ready for training we should merge this now.

PR #2556, opened 2023-03-10
feat(doc) adding workflow to deploy docs

jasonmosley:
The purpose of this workflow is to deploy the document site as features are merged into develop. This will allow users to see the current docs and docs for new beta features ahead of releasing them.


jasonmosley:
@crherman7 We will need to update the config

```
export default defineConfig({
  site: 'https://brandingbrand.github.io',
  base: '/flagship',
})
```

jasonmosley:
@xero I could have used the ` withastro/action@v0` but didn't know if we had a policy about using actions not maintained by github? If we can use astros action it would simplify the workflow.

xero:
> @xero if we had a policy about using actions not maintained by GitHub & if we can use Astros actions

yes you can, but please pin the hash not the tag

jasonmosley:
@crherman7  can you test this to ensure I covered all the URL changes needed?

docs/src/lib/config.ts
```diff
@@ -1,49 +1,58 @@
...
       text: "plugin-google-signin",
-      link: "/en/packages/plugins/google-signin",
+      link: "/flagship/en/packages/plugins/google-signin",
     },
-    { text: "plugin-leanplum", link: "/en/packages/plugins/leanplum" },
+    { text: "plugin-leanplum", link: "/flagship/en/packages/plugins/leanplum" },
     {
       text: "plugin-native-navigation",
-      link: "/en/packages/plugins/native-navigation",
+      link: "en/packages/plugins/native-navigation",
```
crherman7, line 32: `/flagship/`

docs/src/lib/config.ts
```diff
@@ -1,49 +1,58 @@
...
     },
     {
       text: "plugin-target-application",
-      link: "/en/packages/plugins/target-application",
+      link: "/flagship/en/packages/plugins/target-application",
     },
     {
       text: "plugin-target-extension",
-      link: "/en/packages/plugins/target-extension",
+      link: "en/packages/plugins/target-extension",
```
crherman7, line 48: `/flagship/`

crherman7 (changes requested):
Also look for paths in the `.md` and `.mdx` files - they are also missing the prefix

crherman7:
```
[dev] Requests for items in your public folder must also include your base. /flagship-code--dark.svg should be /flagship/flagship-code--dark.svg. Omitting the base will break in production.
```

we also need to prepend every public asset with `/flagship/`

jasonmosley:
This ready for review

xero (approved):
👁️👁️👍

crherman7 (approved):

PR #2560, opened 2023-03-10
fix(fs12): plugin-google-signing peer dependency

crherman7:


jasonmosley:
This is the same as: https://github.com/brandingbrand/flagship/pull/2559

PR #2564, opened 2023-03-13
fix(fs12): logger restoration

crherman7:


packages/core/src/executors/init/pre/logger.ts
```diff
@@ -14,19 +14,14 @@ export const execute = withSummary(
...
-          "log.level = 'error'\n"
-        );
+      try {
+        const log = require(path.project.resolve("node_modules", "npmlog"));
+        log.level = "silent";
+      } catch (e) {
+        //
       }
+
+      writable.redirect();
```
crherman7, line 24: reverse these

PR #2566, opened 2023-03-14
fix(fs12): devDep and peerDep code-core

crherman7:


crherman7:
> I need more details on why this change is needed and what issues it will be solving.

What was found from Cameron when doing his local plugin is that if you did it the typical way that we had suggested by having a local plugin install code-core as a dependency, which it is in the end it is still a required dependency, the idea is that if you were to install dependencies in that local plugin folder it would then have it's own set of node_modules. This typically is not an issue but due to us having the summary that is a single instance, the `withSummary` would then be in a separate instance and therefore the summary for that plugin would go unreported. 

This brings up that we actually need a single instance only installed which leads to peer dependencies. This wasn't an option at first due to issue with changesets which I was able to get a workaround for. This would also be an issue if people had multiple versions of the core package installed. This is the typical way that a repo like expo would do development.

jasonmosley (approved):

devs9 (approved):
👍

PR #2580, opened 2023-03-22
chore(fs12): husky pre commit formatting

crherman7:


jasonmosley (approved):
100% need this thanks

PR #2584, opened 2023-03-22
feat(fs12): dynamic nohoist path resolution

crherman7:
This alleviates `**` nohoist declaration. Hoisting has now been reduced to `react` and `react-native`.

Technically, there could be a zero nohoist policy if we fix the relative paths in the `react-native` typescript template. This is possible through using Node's `require.resolve`.

`require.resolve` is beneficial when downstream files other than the classic `require` is necessary i.e. `react-native-navigation` autolink script, etc. Instead of requiring a specific path, we can use `require.resolve` to dynamically retrieve the path starting at the `process.cwd()`.


packages/core/src/executors/init/platform/plugins.ts
```diff
@@ -11,10 +11,7 @@ export const execute = (options: InitOptions, config: Config) => {
     platform: "ios" | "android"
   ) => {
     for (const installedPlugin of installedPlugins) {
-      const bundle = require(path.project.resolve(
-        "node_modules",
-        installedPlugin
-      ));
+      const bundle = require(path.hoist.resolve(installedPlugin));
```
crherman7, line 14: NIT: this was originally an error because of the nohoist policy - without the nohoist policy with respect to all packages i.e. `**` now all the symlinked monorepo packages are on the same level. Previously this would fail the local plugin. The `path.hoist.resolve` solved this because it would dynamically look up the tree from the current ctx.

After thinking about the `path.hoist.resolve` we may want to have a safe resolve on that because if they don't expose the `package.json` via the `exports` if they are using `exports` it will fail packages.

jasonmosley (approved):

PR #2589, opened 2023-03-31
fix(fs12): plugin-leanplum imports

crherman7:
This does bring up some pain points in not being able to test full implementation of plugins/libraries that require some kind of higher level configuration i.e. firebase.

This does bring up some pain points about having a raw template for testing an ejs template for configuration. We may think about how to merge the 2 ideas. Something for a later time to talk about.


jasonmosley (approved):
agree

PR #2590, opened 2023-03-31
fix(fs12): move process to readline tty commands

crherman7:


jasonmosley (approved):
Thanks. This will allow people to run this in a CI/CD without setting the verbose flag since these two methods require `process.stdout` to be a TTY

PR #2650, opened 2024-03-11
feat(fastlane): update fastlane actions

crherman7:
## Describe your changes

Currently we are not utilizing fastlane to it's fullest potential. We can take advantage of multiple available fastlane actions that install profiles, certs, etc. Additionally on android we do not need to utilize the keys script to update gradle properties to link signing configuration. We can instead update it directly through this plugin and utilize environment variables directly in the build.gradle.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## Test Plan

Updated ejs Fastfile templates. Added unit tests.

1. yarn install
2. yarn build:packages
3. yarn test


NickBurkhartBB:
@crherman7 tests are failing, can you also update the description of the PR to not just be the template and have it be a description meaningful to the PR

packages/plugin-fastlane/src/index.ts
```diff
@@ -47,6 +108,29 @@ eval_gemfile(plugins_path) if File.exist?(plugins_path)
...
+
+    await fsk.update(
+      pathk.project.resolve("android", "app", "build.gradle"),
+      /(signingConfigs[.\s\S]+?release \{)[.\s\S]+?\}/m,
+      `$1
+            storeFile file('release.keystore')
+            storePassword System.getenv("STORE_PASSWORD")
+            keyAlias '${config.android.signing.keyAlias}'
+            keyPassword System.getenv("KEY_PASSWORD")
+        }
```
NickBurkhartBB, line 131: @crherman7 I am ending up with an invalid app/build.gradle in example app project after init, adding the closing bracket here causes a duplicate one cause there currently is an[ if statement](https://github.com/brandingbrand/flagship/blob/main/packages/core/assets/template/android/app/build.gradle#L120) inside release signingConfig and we remove the if block but the closing bracket from release still exists

NickBurkhartBB (changes requested):

packages/plugin-fastlane/src/index.ts
crherman7: good catch updated

NickBurkhartBB (approved):

PR #2667, opened 2024-04-09
chore(version): v13 alpha releases

crherman7:
## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

- [ ] A self-review of my code has been completed


NickBurkhartBB:
@crherman7 I don't think that all the packages changed, I think only code-cli, code-cli-kit, code-plugin-splash-screen, and code-plugin-target-extension need the patch bump

crherman7:
> @crherman7 I don't think that all the packages changed, I think only code-cli, code-cli-kit, code-plugin-splash-screen, and code-plugin-target-extension need the patch bump

correct, I agree - but I thought for alpha versions it would be easier to keep everything in line, unless you disagree still then I will open new pr.

NickBurkhartBB:
@crherman7 I don't think we need to keep them all in sync and shouldn't be hard to track, lets just publish new alphas for the packages we changed

crherman7:
> @crherman7 I don't think we need to keep them all in sync and shouldn't be hard to track, lets just publish new alphas for the packages we changed

ok we are aligned - we need to include changesets on each PR from now on.

PR #2700, opened 2024-05-17
feat(env): ENG-5304 support fsapp <11

crherman7:
## Describe your changes

Env access is linked through `@brandingbrand/fsapp` with the `project_env_index.js` file. In fsapp v10 it's in the root directory whereas in v11 it's in the src directory. We need to conditionalize where we read that file from based on fsapp version.

Require the package.json of `@brandingbrand/fsapp` and determine which path we can use for the `project_env_index.js`.

[Ticket](https://brandingbrand.atlassian.net/browse/ENG-5304)

## Type of change

## Test Plan

1. `yarn install`
2. `yarn build`
3. `cd apps/example`
4. `yarn flagship-code prebuild --build internal --env prod`
5. Verify cli runs successfully

## Demo

<details>
  <summary>cli</summary>

Screenshot 2024-05-17 at 3 25 15 PM
</details>

### DIFF
--- a/packages/cli/src/actions/env.ts
+++ b/packages/cli/src/actions/env.ts
@@ -113,10 +113,13 @@ export default defineAction(
     let projectEnvIndexPath;
 
     if (semver.satisfies(coercedVersion, "<11")) {
-      projectEnvIndexPath = require.resolve(
-        "@brandingbrand/fsapp/project_env_index.js",
+      // project_env_index.js doesn't exist in @brandingbrand/fsapp <v11 - it is assumed to be written to the root
+      // directory of the package. We can get the path based on the package.json and resolve to parent directory with
+      // the project_env_index.js identifier.
+      projectEnvIndexPath = path.resolve(require.resolve(
+        "@brandingbrand/fsapp/package.json",
         { paths: [process.cwd()] }
-      );
+      ), "..", "project_env_index.js");
     }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/actions/env.ts
```diff
@@ -94,12 +95,40 @@ export default defineAction(
...
     // There is a chance this could throw an error, this is fine, still even though we checked the dependencies object already
-    const projectEnvIndexPath = require.resolve(
-      "@brandingbrand/fsapp/src/project_env_index.js",
-      { paths: [process.cwd()] }
-    );
+    let projectEnvIndexPath;
+
+    if (semver.satisfies(coercedVersion, "<11")) {
+      projectEnvIndexPath = require.resolve(
+        "@brandingbrand/fsapp/project_env_index.js",
```
NickBurkhartBB, line 117: @crherman7 when testing, it looks like version 10.x.x.x doesn't have a placeholder file, so this resolutions fails. Error: Cannot find module '@brandingbrand/fsapp/project_env_index.js'.

NickBurkhartBB (changes requested):

packages/cli/src/actions/env.ts
crherman7: great catch! what an interesting tight coupling.

PR #2718, opened 2024-06-12
fix(adr): disable network security config by default

crherman7:
## Describe your changes

When the network security configuration is added to the `AndroidManifest.xml` by default with an empty configuration it can interfere with the metro packager when developing.

Only update the `AndroidManifest.xml` with the network security configuration if someone utilizes the the parser. Note, the developer will be responsible for adding the local environments in non-release mode via their plugin.

## Type of change

## Test Plan

- `yarn test`

### DIFF
--- a/packages/cli/src/transformers/android/android-manifest-xml.ts
+++ b/packages/cli/src/transformers/android/android-manifest-xml.ts
@@ -105,25 +105,6 @@ export default defineTransformer<Transforms<AndroidManifestXML, void>>({
 
       mainActivity.$['android:screenOrientation'] = orientation;
     },
-
-    /**
-     * Function that applies the network security config to the AndroidManifest.xml file.
-     * @param xml The AndroidManifestXML object representing the contents of the AndroidManifest.xml file.
-     * @param config The build configuration containing Android-specific manifest options.
-     */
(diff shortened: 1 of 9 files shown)


packages/cli/src/transformers/android/android-manifest-xml.ts
```diff
@@ -105,25 +105,6 @@ export default defineTransformer<Transforms<AndroidManifestXML, void>>({
...
-      const mainApplication = xml.manifest.application
-        ?.find((it) => it.$["android:name"] === ".MainApplication")
-
-      if (!mainApplication) {
-        throw new Error(
-          "[AndroidManifestTransformer]: cannot set network security configuration because .MainApplication not found"
-        );
-      }
-
-      mainApplication.$["android:networkSecurityConfig"] = "@xml/network_security_config"
```
NickBurkhartBB, line 125: @crherman7 should we also not add the empty network_security_config file since it is unused?

crherman7: updated to dynamically write the file if it doesn't exist when attempting to use the parser

NickBurkhartBB (approved):

PR #2763, opened 2024-10-29
fix: ENG-5789 - Version pin xcproj and cocoapods for CI

UncleClapton:

[ENG-5789](https://brandingbrand.atlassian.net/browse/ENG-5789)

## Type of change

[ENG-5789]: https://brandingbrand.atlassian.net/browse/ENG-5789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/cli/src/actions/packagers.ts
```diff
@@ -98,7 +98,7 @@ export default defineAction(async (): Promise<void> => {
     try {
       for await (const line of execa({
         cwd: path.project.resolve('ios'),
-      })`pod install`) {
+      })(isCI ? 'bundle exec pod install' : 'pod install')) {
```
NickBurkhartBB, line 101: @UncleClapton I am getting error is I run with `CI=1 yarn run prebuild -e dev -b internal --verbose` in the example app. I believe it is how handles a template literal versus if it receives a string. I think with string this way you need to specify like (command, [args]), (isCI ? 'bundle' : 'pod', isCI ? ['exec', 'pod', 'install'] : ['install'])

NickBurkhartBB (approved):

packages/cli/src/actions/packagers.ts
NickBurkhartBB: maybe its' something with node versions?

PR #2813, opened 2025-07-29
feat: ENG-6217 - Dynamic app ENV provider support

UncleClapton:
## Describe your changes

* Dynamically support either `@brandingbrand/fsapp` or the upcoming `@brandingbrand/code-app-env` in `code-plugin-env` based on package dependencies
* Update code plugin template to include common function, as well as an improved explanation on plugins.

[ENG-6217](https://brandingbrand.atlassian.net/browse/ENG-6217)

## Type of change

## Test Plan

1. Setup a project using FSApp, and ensure all FSCode packages are resolved to local file versions
2. Run `yarn flagship-code prebuild` and ensure `code-plugin-env` injects your expected envs into `node_modules/@brandingbrand/fsapp/src/project_env_index.js`
3. Add `@brandingbrand/code-app-env` to your project's dependencies array. 
    * The version specifier does not matter and only it's existence in `package.json` is necessary to test the plugin. No need to run `yarn install`
4. Run `yarn flagship-code prebuild` again. Expect a warning with both packages "installed".
5. Confirm:
    1. `.flagshipappenvrc` has been generated in project top level directory.
    2. Check `./android/app/src/main/res/values/strings.xml` and confirm `flagship_env` and `flagship_dev_menu` is set
    3. Check `./ios/app/Info.plist` and confirm `FlagshipEnv` and `FlagshipDevMenu` are set.

[ENG-6217]: https://brandingbrand.atlassian.net/browse/ENG-6217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/plugin-env/src/index.ts
```diff
@@ -1,193 +1,71 @@
...
+    ...Object.keys(pkgJson.dependencies || {}),
+    ...Object.keys(pkgJson.devDependencies || {}),
+  ];
+};
 
-          return {name: envName, content: mod.default};
-        }),
-      );
-    };
+const projectDeps = loadProjectDependencyList();
```
NickBurkhartBB, line 23: this call can throw an error, but since it is now at the top level, it could be an uncontrolled exception. we may want to rethink the return type and maybe handle error at the individual function level if we are trying to load at the top level.

UncleClapton: The error would be caught while the plugin is loading in, but that would definitely create an error that's harder to trace. Moved to Common function instead.

packages/plugin-env/src/index.ts
```diff
@@ -1,193 +1,72 @@
...
+  common: async (build, options) => {
+    logger.info('Configuring runtime environment provider.');
 
-          return {name: envName, content: mod.default};
-        }),
-      );
-    };
+    // Determine which plugin to use based on the presence of a supported package in the project's package.json
+    const projectDeps = loadProjectDependencyList();
+    activePlugins = Object.values(plugins).filter(it =>
```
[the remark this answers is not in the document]
UncleClapton: I agree it's a bit over-built for only two supported packages, but sticking to an iteration approach simplifies the user feedback logic, and avoids needless rechecking across the 3 plugin stages.

I've removed the custom plugin typing because that's definitely over-complicating matters when we could just use a mapping object.

PR #2835, opened 2026-02-02
fix: defaults for newArchEnabled driven by template

NickBurkhartBB:
## Describe your changes
remove the logic in 0.77 grade-properties.ts to fallback to the logic already in the 0.72 which should be the expected
if a user defines the value it is used, if not defined it just uses the values from the version of the template you are using

https://brandingbrand.atlassian.net/browse/ENG-6287

## Type of change

## Test Plan
* run prebuild in example app and check android/gradle.properties
* ER: it shows newArchEnabled=true as that is the template default
* update build.internal.ts in the example app to a value and re-run prebuild
* ER: it shows newArchEnabled={value you set}


UncleClapton (changes requested):
I can confirm this fixes the issue for android, but I think we should update iOS as well.

there's a couple of issues with the current iOS transform.
- it only exists on RN77+. The method in which we enable new-arch for iOS exists as far back as RN72, however, so we could move this to the RN72 podfile transforms.
- the current iOS transform behaves like the android RN77 transform, setting new arch to `false` if the property isn't in the build config. We should update this to behave like Android's transform.

I also think we should add transform overrides for RN82+ to completely ignore `newArchEnabled` in the build config, since this setting is a no-op for both platforms in RN82+.

NickBurkhartBB:
@UncleClapton good callouts. I updated iOS to have the same undefined logic and moved back to 0.72 version. I also added 0.82 transform to add a warning to notify user of safe cleanup and have its value ignored.

.changeset/dirty-jobs-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@brandingbrand/code-plugin-transform-template": patch
+---
+
+fix Android defaults for newArchEnabled to be tied to template
```
UncleClapton: could we update the changeset to reflect the latest changes?

UncleClapton (changes requested):

.changeset/dirty-jobs-argue.md
NickBurkhartBB: removed Android qualifier as it applies to both platforms now. added line about 0.82+ behavior

UncleClapton (approved):
LGTM!

PR #2881, opened 2026-07-23
feat: harden react-native version adoption tooling and CI

deemaabdallah:
## Summary
- `scripts/add-rn-version` now resolves the latest stable patch for the requested minor from the registry and initializes with that exact version. Minor-only versions relied on `0.XX-stable` dist-tags, which are no longer consistently published upstream. Unknown or invalid minors exit with a clear error instead of continuing.
- The version codemods also update the root `package.json` React Native pin (adding an older version never downgrades it). App workspaces are aligned from the dependency profile via `flagship-code align-deps --fix`, keeping the profile as the single source of truth for app-level versions.
- New tests in `plugin-verify-dependencies` assert that every dependency profile pins `react-native` and the `@react-native/*` packages to its own version, and that profiles cover exactly the set of committed templates.
- The Jest template environment's default version can be overridden via `FLAGSHIP_CODE_TEST_RN_VERSION`; CI runs the plugin suites against every committed template version (the version list derives from `packages/templates/react-native/`).
- New `PR Compile` workflow builds the generated Android and iOS projects when templates or the example app change. Dependabot PRs skip it.

## Verification
- `yarn build`, `yarn lint`, `yarn test` green.
- The profile assertion fails when a profile pins a version other than its own (verified by mutation, then reverted).
- Full test suite green with `FLAGSHIP_CODE_TEST_RN_VERSION` set to 0.75, 0.79, and 0.83.
- `./scripts/add-rn-version 0.84` resolves 0.84.1 and completes end to end (output discarded); `./scripts/add-rn-version 0.99` exits with an invalid-version error.
- `align-deps --profile <version> --fix` verified to rewrite an app's React Native pins from the profile, including devDependency routing.
- Android and iOS compile jobs green on the generated example app.

Note: `plugin-native-navigation`'s default suite is pinned to the 0.72 template it targets.

### DIFF
--- a/.github/workflows/pr-compile.yml
+++ b/.github/workflows/pr-compile.yml
@@ -1,10 +1,12 @@
 name: PR Compile
 
 on:
+  workflow_dispatch:
   pull_request:
     paths:
       - 'packages/templates/**'
       - 'apps/example/**'
+      - '.github/workflows/pr-compile.yml'
 
 jobs:
   android:
--- a/.github/workflows/pr-compile.yml
+++ b/.github/workflows/pr-compile.yml
@@ -1,0 +1,61 @@
+name: PR Compile
+
+on:
+  pull_request:
+    paths:
+      - 'packages/templates/**'
+      - 'apps/example/**'
+
+jobs:
+  android:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Use Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8
+        with:
+          node-version: 20
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@f26937343756480a8cb3ae1f623b9c8d89ed6984
+        with:
+          ruby-version: '3.1'
+      - name: Setup Java
+        uses: actions/setup-java@6a0805fcefea3d4657a47ac4c165951e33482018
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - name: Install dependencies
+        run: yarn
+      - name: Build packages
+        run: yarn build
+      - name: Initialize react native app
+        run: yarn workspace @brandingbrand/code-example prebuild --build internal --env prod --platform android --verbose
+      - name: Compile Android app
+        run: ./gradlew assembleDebug
+        working-directory: apps/example/android
+
+  ios:
+    runs-on: macos-15
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+      - name: Use Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8
+        with:
+          node-version: 20
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@f26937343756480a8cb3ae1f623b9c8d89ed6984
+        with:
+          ruby-version: '3.1'
+      - name: Install dependencies
+        run: yarn
+      - name: Build packages
+        run: yarn build
+      - name: Initialize react native app
+        run: yarn workspace @brandingbrand/code-example prebuild --build internal --env prod --platform ios --verbose
+      - name: Compile iOS app
+        run: xcodebuild -workspace app.xcworkspace -scheme app -configuration Debug -sdk iphonesimulator -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO build
+        working-directory: apps/example/ios
--- a/.github/workflows/pr-compile.yml
+++ b/.github/workflows/pr-compile.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   android:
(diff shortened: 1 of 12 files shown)


.github/workflows/pr-compile.yml
```diff
@@ -0,0 +1,61 @@
+name: PR Compile
```
NickBurkhartBB, line 1: personally, I think we shouldn't add this compile on pull_request. These can take a long time and dependabot updates will each trigger them. for now, we have left this verification for the reviewer as needed, but if we feel strongly about adding these in, we can.

apps/example/flagship-code.config.ts
```diff
@@ -44,18 +44,20 @@ export default defineConfig({
...
+   * - react-navigation: Configures native pieces for React Navigation
+   * - monorepo: Provides example implementations
    * @type {string[]}
    */
   plugins: [
     '@brandingbrand/code-plugin-asset',
     '@brandingbrand/code-plugin-app-icon',
     '@brandingbrand/code-plugin-splash-screen',
+    '@brandingbrand/code-plugin-permissions',
+    '@brandingbrand/code-plugin-react-navigation',
```
NickBurkhartBB: example doesn't have these plugins peer dependencies setup in package.json for these two or have any code in the example the tie to these dependencies. If we are expanding the example with these code plugins, we also need to add the other dependencies and update the example to fully test.

scripts/add-rn-version
```diff
@@ -44,6 +44,18 @@ if ! [[ "$RN_VERSION" =~ ^0\.[0-9]{2}$ ]]; then
 fi
 
 
+# Resolve the latest stable patch release for the requested minor version.
+# The CLI resolves a minor-only version through the `0.XX-stable` npm dist-tag,
+# which upstream no longer publishes reliably, so we resolve the exact patch
+# version ourselves and pass that to `init`.
+EXACT_VERSION=$(npm view "react-native@${RN_VERSION}.x" version --json 2>/dev/null \
```
NickBurkhartBB, line 51: npm responds with error object for invalid version. so running `./scripts/add-rn-version 0.99` causes it to resolve an error object, then continues instead of exiting as invalid version. We need to add additional safety to the parsing of the response from npm. Possible fix could be like treat anything that isn't an array or a bare string as empty: `console.log(Array.isArray(v) ? (v[v.length-1] ?? '') : (typeof v === 'string' ? v : ''))`

NickBurkhartBB (changes requested):

apps/example/flagship-code.config.ts
deemaabdallah: yes planning to expand on the example app. I will remove and can be added in relevant PR.

.github/workflows/pr-compile.yml
deemaabdallah: I can add a dependabot skip to both jobs so dep bumps never trigger this. With that said, I would like to keep it on pull_request since it only fires on template/example changes anyway and helps us catch failures at compile. On its first run on this PR, it caught the `react-native-screens` codegen break that i`nstall/lint/test/prebuild `all missed. 

Since the goal is to be doing RN version bumps regularly, it makes those PRs reviewable without building both platforms locally. I can take it out completely instead if that works better or let me know if you suggest another route.

NickBurkhartBB: alright, we can keep it.

scripts/add-rn-version-codemods.ts
```diff
@@ -276,6 +276,81 @@ async function updateDependencyProfilesIndex() {
...
+    path.resolve(process.cwd(), './apps/example/package.json'),
+    'devDependencies',
+    [
+      '@react-native/babel-preset',
+      '@react-native/eslint-config',
+      '@react-native/metro-config',
+      '@react-native/typescript-config',
+    ],
+  );
+}
```
UncleClapton: Since FSCode can already perform the version bumps in the example app through `align-deps --fix`, I'm curious if there's a way we could incorporate that at some later stage, instead of bumping them in this script.

I considered updating the example app as a part of these scripts, but decided to leave it out for two reasons: 

First is that the example app is a good platform for testing the `align-deps` command. After FSCode is updated, we want a way to ensure all packages will be updated as expected.

Second is that there are other packages we need to update when the example is upgraded, and this script lacks the context to make those updates. After `align-deps` is fully configured for the new version, it would be a better tool for this task

I am primarily concerned about the CLI packages, as well as RNScreens and React Navigation if we're adding that in now. All three have strict RN version compatibility requirements, and are on independent release cycles.

UncleClapton (commented):
I like the direction this is going in. I'm just not sure if the codemod script should also be concerned with anything within the example app. More details below

As an aside, I think we should start tracking RNScreens as an alignment dependency in `verify-dependencies`. This package has some strict version alignment requirements https://github.com/software-mansion/react-native-screens#support-for-fabric

scripts/add-rn-version-codemods.ts
deemaabdallah: love this! thanks @UncleClapton I will test locally and fix the PR accordingly.

NickBurkhartBB (approved):

UncleClapton (approved):
lgtm!

PR #2885, opened 2026-07-24
feat(example): exercise permissions and react-navigation plugins

deemaabdallah:
## Summary
- Enables `@brandingbrand/code-plugin-permissions` and `@brandingbrand/code-plugin-react-navigation` in the example app, with their peer dependencies (`react-native-permissions ^4.1.5` per the plugin's peer range, `react-native-screens ~4.16.0` aligned with the app's React Native version) and app code that exercises both at runtime.
- Adds a native-stack navigation setup (`@react-navigation/native` + `@react-navigation/native-stack`): Home and Permissions screens.
- The Permissions screen checks and requests the camera permission declared through the permissions plugin, so the plugin's native output is used by real code rather than only generated.

## Verification
- `yarn build`, `yarn lint`, `yarn test` green.
- Android prebuild green; the camera permission and the predictive-back opt-out are present in the generated manifest.
- iOS prebuild green; `NSCameraUsageDescription` present in the generated Info.plist; `pod install` clean (85 pods, RNPermissions and RNScreens codegen processed).

### DIFF
--- a/apps/example/package.json
+++ b/apps/example/package.json
@@ -15,9 +15,13 @@
     "@brandingbrand/code-app-env": "*",
     "@brandingbrand/react-native-app-restart": "^0.5.0",
     "@react-native-async-storage/async-storage": "^2.2.0",
+    "@react-navigation/native": "^7.3.14",
+    "@react-navigation/native-stack": "^7.18.6",
     "react": "^19.2.0",
     "react-native": "^0.83.0",
-    "react-native-safe-area-context": "^5.6.2"
+    "react-native-permissions": "^4.1.5",
+    "react-native-safe-area-context": "^5.6.2",
+    "react-native-screens": "~4.16.0"
   },
   "devDependencies": {
     "@babel/core": "^7.28.5",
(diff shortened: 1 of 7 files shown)


apps/example/package.json
```diff
@@ -15,9 +15,13 @@
     "@brandingbrand/code-app-env": "*",
     "@brandingbrand/react-native-app-restart": "^0.5.0",
     "@react-native-async-storage/async-storage": "^2.2.0",
+    "@react-navigation/native": "^7.3.14",
+    "@react-navigation/native-stack": "^7.18.6",
     "react": "^19.2.0",
     "react-native": "^0.83.0",
-    "react-native-safe-area-context": "^5.6.2"
+    "react-native-permissions": "^4.1.5",
```
NickBurkhartBB, line 22: this should probably be version 5, which actually show cases we probably need an update to the plugin to verify on 5 and update peer dependencies to allow it before updating the example. Maybe we just do React Navigation in example now, then a later update to include permissions after plugin update?

deemaabdallah: @NickBurkhartBB agreed but this should be done in a separate ticket  i kept permissions at v4 on purpose for now, since that's what code-plugin-permissions verifies and peer-deps today. Did you see any issues with current state?

NickBurkhartBB: @deemaabdallah I stopped from further exploration as I felt this would block it. I can try to run through more tests and review on this version.

NickBurkhartBB (approved):
@deemaabdallah version 4.1.5 of permissions still works with 0.82 so the example changes work. 
Note: this may add delays in RN updates as we now are reliant on more pieces outside RN releases.

PR #2886, opened 2026-07-24
docs: describe react-native support as policy rather than enumerating versions

deemaabdallah:
## Summary
- The getting-started guide stated Flagship Code v13 with support for React Native 0.72 and 0.73; current reality is v14 with a wider supported range. Rather than updating one enumeration to another that will also go stale, the supported-versions statements now describe how support works: a range of recent minors per major release, new releases added shortly after they ship, older versions retired in major releases, and the CLI listing supported versions when it encounters an unrecognized one.
- Code examples updated to current versions (react ^19.2.0, react-native ^0.83.0).

## Verification
- Docs site builds clean (`yarn workspace @brandingbrand/code-docs build`).

### DIFF
--- a/apps/docs/src/content/docs/guides/getting-started.mdx
+++ b/apps/docs/src/content/docs/guides/getting-started.mdx
@@ -10,25 +10,25 @@ Flagship Code is a comprehensive Configuration as Code (CaC) toolkit designed to
 
 In your `package.json` verify you are using matching the versions of `react-native` and `react` packages noted below.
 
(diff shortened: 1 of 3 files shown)


NickBurkhartBB:
@deemaabdallah looks like we have some other instances that are stale in the docs like https://github.com/brandingbrand/flagship/blob/develop/apps/docs/src/content/docs/overview/objectives.md?plain=1#L17

deemaabdallah:
@NickBurkhartBB this should be ready to review once cI passes

NickBurkhartBB (approved):

PR #2887, opened 2026-07-27
ci: add weekly RN release-watch workflow

christophermason:
Watches npm for new React Native minors and files one `rn-adoption`-labeled tracking issue per pending minor above the highest adopted, ascending and deduped against existing tracking issues (a jump straight to a new major is caught the same way). Adoption itself stays manual (`yarn add-rn-version <version>`); this only detects and files the issue. Uses the stock `GITHUB_TOKEN` (`issues: write` only).

The logic works as follows:

- Each filed issue points to the published maintainer guide (`add-rn-version`) instead of an inline recipe, and links the upstream template diff (`rn-diff-purge`) for the target minor.
- npm-deprecated dot-zero releases (e.g. `react-native@1000.0.0`, accidentally published in 2024) are excluded so a bogus version can't be picked up as pending.
- Detection logic (pending-minor computation, publish check, deprecated-version exclusion, duplicate-issue check) lives in `scripts/rn-release-watch.mjs` as pure, dependency-injected functions callable outside Actions.
- Unit tests in `scripts/rn-release-watch.test.mjs` (`node --test`), including an end-to-end test that runs the watcher against mocked npm/GitHub responses and asserts the exact issue title/body it would file.
- A `test` job runs on every PR touching these files; the `watch` job itself runs weekly (Thursdays 13:00 UTC) or on manual dispatch (with a `dry_run` input).
- No maturity wait: a freshly-published (0-day-old) dot-zero release is filed immediately.

## Test plan
- [x] `node --test scripts/rn-release-watch.test.mjs`
- [x] `actionlint .github/workflows/rn-release-watch.yml`
- [x] Dry run against real npm/GitHub data (`DRY_RUN=true`)
- [ ] First scheduled/manual run in CI

### DIFF
--- a/scripts/rn-release-watch.mjs
+++ b/scripts/rn-release-watch.mjs
@@ -1,0 +1,227 @@
+#!/usr/bin/env node
+// Watches npm for every React Native minor whose dot-zero release has been
+// published but isn't adopted yet, and files one `rn-adoption` tracking
+// issue per pending minor (ascending, deduped against existing issues).
+// Adoption itself stays manual (`yarn add-rn-version <version>`) -- this
+// only detects and files the issues.
+
+import {readdirSync} from 'node:fs';
+
+const TEMPLATE_DIR = new URL(
+  '../packages/templates/react-native/',
+  import.meta.url,
+);
+
+const MAINTAINER_GUIDE_URL =
+  'https://brandingbrand.github.io/flagship/maintain/add-rn-version/';
+
+/**
+ * Given the directory names under packages/templates/react-native, return
+ * the highest adopted React Native minor.
+ */
+export function computeHighestAdopted(templateDirNames) {
+  const adopted = templateDirNames
+    .filter(name => /^\d+\.\d+$/.test(name))
+    .map(name => name.split('.').map(Number))
+    .sort((a, b) => a[0] - b[0] || a[1] - b[1]);
+
+  if (adopted.length === 0) {
+    throw new Error('No adopted react-native template versions found.');
+  }
+
+  const [major, minor] = adopted[adopted.length - 1];
+  return {major, minor, label: `${major}.${minor}`};
+}
+
+/**
+ * Given the highest adopted minor and the npm packument's `time` map, return
+ * every published dot-zero minor above it, ascending. Compares (major,
+ * minor) tuples rather than assuming a fixed major, so a jump straight to a
+ * new major (e.g. 1.0) is caught instead of silently missed.
+ *
+ * `npmVersions` is the packument's `versions` map -- entries npm marks
+ * `deprecated` are excluded. react-native@1000.0.0 (published 2024-01-31,
+ * "accidentally published and should not be used in production") is real
+ * data that would otherwise be picked up as the "highest" pending minor.
+ */
+export function computePendingMinors({highest, npmTime, npmVersions}) {
+  const pending = [];
+  for (const [version, releasedAt] of Object.entries(npmTime)) {
+    const match = /^(\d+)\.(\d+)\.0$/.exec(version);
+    if (!match) {
+      continue;
+    }
+    if (npmVersions[version]?.deprecated) {
+      continue;
+    }
+    const major = Number(match[1]);
+    const minor = Number(match[2]);
+    if (
+      major > highest.major ||
+      (major === highest.major && minor > highest.minor)
+    ) {
+      pending.push({
+        major,
+        minor,
+        label: `${major}.${minor}`,
+        version,
+        releasedAt,
+      });
+    }
+  }
+  return pending.sort((a, b) => a.major - b.major || a.minor - b.minor);
+}
+
+/**
+ * Pure date-math, for logging only -- there is no maturity wait gating
+ * whether a pending minor is eligible.
+ */
+export function evaluateRelease({releasedAt, now}) {
+  const ageDays = Math.floor(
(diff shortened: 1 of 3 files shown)


scripts/rn-release-watch.mjs
```diff
@@ -0,0 +1,186 @@
...
+  ageDays,
+  highest,
+  stabilityWindowDays,
+}) {
+  return [
+    `React Native ${dotZeroVersion} shipped on ${releasedAt} and has now cleared the ${stabilityWindowDays}-day stability window (${ageDays} days old).`,
+    '',
+    `This repo currently ships native templates through ${highest} (\`packages/templates/react-native\`); ${nextMinor} is not yet supported.`,
+    '',
+    'This issue only *tracks* adoption -- it is not automated (spec non-goals). To adopt:',
```
NickBurkhartBB: what is this spec non-goals being referenced?

scripts/rn-release-watch.mjs
```diff
@@ -0,0 +1,186 @@
...
+    '',
+    `1. \`yarn add-rn-version ${nextMinor}\` to generate the native templates and run the codemods (\`scripts/add-rn-version\`).`,
+    '2. Review the generated diff, update changelogs, and open a PR through the usual process.',
```
NickBurkhartBB: there is more to the process, we need to update example and verify it actually works.

scripts/rn-release-watch.mjs
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env node
+// Watches npm for a newly-stable React Native minor and files an `rn-adoption`
+// tracking issue once it clears a stability window. Adoption itself stays
+// manual (`yarn add-rn-version <version>`) per spec's non-goals -- this only
+// detects and files the issue.
+
+import {readdirSync} from 'node:fs';
+
+export const STABILITY_WINDOW_DAYS = 14;
```
NickBurkhartBB: as we are just creating a issue, I don't think we need to wait 14 days. We should create the issue whenever the version hits.

NickBurkhartBB (commented):
As GitHub Actions have gh cli already installed, I feel like this can be simplified by writing as a shell script with gh/jq/etc. I think this would shrink down on the noise of the script in setting up all these requests
As this is personal preference and if other people feel strongly about being in Javascript, I'm ok keeping in that

christophermason:
### Live dry-run proof (workflow_dispatch, dry_run=true)

Ran the workflow against the real npm registry with `dry_run=true`, so it detects and renders but files nothing.

Run: https://github.com/brandingbrand/flagship/actions/runs/30372482371 (test job 15/15 green, then the watch job below)

```
DRY_RUN: true
Highest adopted RN minor: 0.83
Pending minors above 0.83: 0.84, 0.85, 0.86
0.84.0 published 2026-02-11T16:04:19.627Z (166 days old).
[dry run] Would file issue "RN 0.84 adoption tracking":
React Native 0.84.0 shipped on 2026-02-11T16:04:19.627Z.

This repo currently ships native templates through 0.83 (`packages/templates/react-native`); 0.84 is not yet supported.

Follow the [Adding React Native Version Support](https://brandingbrand.github.io/flagship/maintain/add-rn-version/) guide for the full adoption process -- this issue only tracks that 0.84 is ready to adopt.

Upstream template diff: https://react-native-community.github.io/upgrade-helper/?from=0.83.0&to=0.84.0

_Filed automatically by `.github/workflows/rn-release-watch.yml`._
0.85.0 published 2026-04-07T17:43:56.156Z (111 days old).
[dry run] Would file issue "RN 0.85 adoption tracking":  (guide + diff from=0.83.0&to=0.85.0)
0.86.0 published 2026-06-09T13:51:59.164Z (49 days old).
[dry run] Would file issue "RN 0.86 adoption tracking":  (guide + diff from=0.83.0&to=0.86.0)
```

Confirms the revised behavior end to end:
- one tracking issue per pending minor (0.84, 0.85, 0.86), ascending, not just the next one
- issue body points to the published maintainer guide, no hand-written recipe
- each body carries the upstream Upgrade Helper template diff
- npm's deprecated `1000.0.0` is excluded, so no bogus RN 1000 issue
- dry_run filed nothing (0 issues created)

Weekly schedule is live in the workflow: `cron: '0 13 * * 4'` (Thursday 13:00 UTC).

NickBurkhartBB (approved):
works in testing for me

scripts/rn-release-watch.mjs
```diff
@@ -0,0 +1,227 @@
...
+} = {}) {
+  const token = process.env.GITHUB_TOKEN;
+  const repository = process.env.GITHUB_REPOSITORY; // "owner/repo"
+  if (!token || !repository) {
+    throw new Error('GITHUB_TOKEN and GITHUB_REPOSITORY must be set.');
+  }
+  const [owner, repo] = repository.split('/');
+
+  const templateDirNames = readTemplateDirNames();
+  const highest = computeHighestAdopted(templateDirNames);
```
UncleClapton, line 145: We've previously had the policy that FSCode supports up to the latest RN version - 1. For example the current version of RN is `0.86`, and so we would only support up to `0.85`. Are we intending to abandon this policy and begin including support for the true latest?

UncleClapton: After speaking internally,

Leave script as-is, I missed that issue generation doesn't automatically lead to a PR generation afterward.

We still want the new version to open an issue, even if we don't act on it immediately.

UncleClapton (approved):

PR #2893, opened 2026-07-28
chore: add agent docs and raise CI to node 22

deemaabdallah:
## Summary

- Add a root `AGENTS.md` orienting coding agents to this repository: how to install and verify, changeset expectations, branch and commit conventions, and the ground rules that are easy to get wrong here.
- Add `.agents/skills/upgrade-rn-version/SKILL.md`, the procedure for adopting a new React Native minor: build-time version detection, harvest, dependency profile, workspace alignment, native compile on both platforms, out-of-bounds paths and stop conditions, a per-PR definition of done, and anti-fake-test rules.
- Raise the Node version CI runs and the repository declares from 20 to 22.

## Why Node 22 is in this change

`@react-native-community/template@0.84.1` declares `engines.node >= 22.11.0`. The example app is generated from that template, so adopting React Native 0.84 on Node 20 fails the compile gate for a reason unrelated to the adoption. The skill added here documents Node 22 as the repository's baseline, so the two belong together rather than describing a state the workflows do not yet match.

All four workflows now run Node 22 and root `engines.node` is `>=22.11.0`. The docs deploy sets it explicitly because the composite action it uses defaults to 20 and only runs on pushes to `main`, so no pull request would have surfaced the mismatch.

## Notes

- No package source changes. The skill and `AGENTS.md` are guidance; the remaining diff is CI configuration and the declared engine.
- The adoption model the skill encodes: one version per pull request, ascending, each proven by a native build of the example app before the next begins.
- `.agents/skills/` is not auto-loaded by any agent client today. `AGENTS.md` is the entry point and points at the skill, which keeps one copy of the procedure rather than a per-client duplicate.
- The skill deliberately avoids restating facts that go stale. Toolchain versions are read from `pr-compile.yml`, the supported-version list from the `packages/templates/react-native/` directory listing, the scoped package set from the previous dependency profile, and the changeset scope from the previous minor's changeset.
- Verification for the skill's own gates is deferred to CI by design. Where a step could be self-reported as passing, the definition of done names the workflow job instead, since a job that runs is harder to conclude about than a checkbox.

## How the procedure was checked

Run end to end in throwaway clones, nine times: three model tiers across three target minors, from a prompt that named neither `AGENTS.md` nor the skill so that locating them was part of the test. One set of runs targeted the lowest pending minor from a base that made it the correct next one. Another targeted a minor two ahead of the newest adopted, to check that the ascending rule holds when a request names the wrong version.

Four defects surfaced that way and all four are fixed here:

- The changeset step named more packages than an adoption touches, so following it declared bumps for packages the work never modifies. The previous minor's changeset is now the reference.
- The premise check ran the harvest against a possibly uninstalled workspace, where the codemods fail for a missing `typescript` module. That reads as broken tooling under a step whose instruction is to stop and report, so an adoption could end before starting.
- A version named in a request was treated as the target, which was enough to skip intervening minors and leave the supported-version list with a gap. The target is now computed from the pending set, with a named version acting as a check on that computation.
- Reconciliation covered template-derived entries and was silent on the rest, which left inherited pins with no template counterpart open to a guessed value, and left a scoped package newly appearing in the template looking like something to adopt rather than an addition.

The machine used had no JDK and no Android SDK, which is a property of that machine and not of CI. What it tested was the response: every run reported it as an environment blocker rather than skipping the build or installing around it. The native builds themselves are only ever proven by `PR Compile`.


AGENTS.md
```diff
@@ -0,0 +1,32 @@
...
+- Read `CONTRIBUTING.md` for setup and commit conventions.
+- Every user-facing change needs a changeset (`.changeset/`) declaring every affected package and its bump type.
+- Read the version's template and the maintainer docs before changing anything native. Native behavior changes between React Native versions; do not assume a pattern from one version holds in another.
+
+## Working conventions
+
+- Open pull requests against `develop`, not `main` (releases cut from `main` via Changesets).
+- Conventional Commits (`type(scope): subject`), one logical change per commit.
+- Pull requests open as drafts; mark ready only when the gates applicable to the change are green.
+- Match existing conventions: profile file shape (`0.83.ts`), workflow style in `.github/workflows/`, commit scopes from recent history.
```
NickBurkhartBB: chance for becoming stale with file reference to a RN version that will at some point go away. might not matter, but might not be needed, or reference top level folder.

.agents/skills/upgrade-rn-version/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+4. **Align and compile the example.** In `apps/example`: `npx flagship-code align-deps --profile 0.XX --fix`, then `yarn install`. Never hand-edit the example's React Native pins; the profile drives them. Then run the native build: `yarn workspace @brandingbrand/code-example prebuild --build internal --env prod --platform android --verbose`, and the iOS equivalent with `pod install`. A version is not adopted until it has compiled as a real app. The plugin test matrix passing alone is not sufficient.
+5. **Ecosystem-native pin breaks.** If the example's compile fails because a non-profile-tracked dependency (screens, gesture-handler, reanimated, safe-area-context, async-storage, webview, community CLI) is incompatible with the new version, bump that one dependency's pin in the example (not the profile mechanism) and record why. The compile is what surfaces such a break instead of shipping a stale pin. Automating these pairings is separate, larger work, out of scope for an adoption PR.
+6. **Full local gate:** `yarn install`, `yarn build`, `yarn lint`, `yarn test`, then the native builds above, all green.
```
NickBurkhartBB: we may want to update `yarn test` to `FLAGSHIP_CODE_TEST_RN_VERSION=0.XX yarn test` so it tests the version it added instead of 0.72.

AGENTS.md
```diff
@@ -0,0 +1,32 @@
...
+
+Orients any coding agent (Claude, Copilot, Cursor, and others) working in this repository. This is a public file. Nothing internal belongs here: no dispatch process, no named reviewers, no links to private threads or docs.
+
+## What this repo is
+
+Flagship Code: a React Native app framework. A CLI, per-version native templates, and dependency-verification tooling for mobile commerce apps.
+
+## Before you touch anything
+
+- Read `CONTRIBUTING.md` for setup and commit conventions.
```
NickBurkhartBB: CONTRIBUTING.md doesn't have setup or commit conventions

.agents/skills/upgrade-rn-version/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. **Premise check.** Confirm `./scripts/add-rn-version` and the CI gates below already exist and work (throwaway-branch dry run). If tooling is missing or broken, that is a blocker on the tooling itself. Stop and report it; do not build or repair tooling as part of an adoption.
+2. **Harvest.** Run `./scripts/add-rn-version 0.XX`. This pulls the upstream `@react-native-community/template` (android and ios only) into `packages/templates/react-native/0.XX/` and codemods: CLI registration (`packages/cli-kit/src/@types/config.ts`, `packages/cli/src/ui/constants/messaging.ts`), a shell dependency profile (`packages/plugin-verify-dependencies/src/profile/0.XX.ts` plus registration in `index.ts`), and the root `package.json` React Native pin.
+3. **Fill the profile.** Copy the inheritance pattern of the newest existing profile: spread the previous minor's profile, then override `react-native` plus the four `@react-native/*` entries to `^0.XX.0`. Source the pairings from the upstream template's own `package.json`. Do not add tracking for packages prior profiles did not track (ecosystem-native libs like screens, gesture-handler, reanimated are not profile-tracked yet; see step 5).
```
NickBurkhartBB: This step 3 might need some work. when running skill for 0.84 it caused it to pickup the caret for react so it used `^19.2.3` instead of `19.2.3` as required for version 0.84 of react-native. It seems to come from `Copy the inheritance pattern of the newest existing profile` as it saw the caret from 0.83 and applied it to 0.84 react version. 
Here is a possible proposal for updating step 3 though it is still up for debate:
```
 3. Fill the profile. Spread the previous minor's profile, then reconcile every template-derived entry against this minor's template. Do not assume an inherited value is still correct.
▎   - Source of truth: this minor's @react-native-community/template package.json (npm pack @react-native-community/template@<patch>). Diff it against the previous minor's — or read the same diff in the upgrade helper — to see what moved.
▎   - What to write: copy the template's version string verbatim, range operator included. Exact pin upstream ("A.B.C") means exact pin in the profile ('A.B.C'); a range upstream ("^A.B.C") stays that range. The operator is part of the pairing — React Native is tested against one patch of its react, and widening it lets a consumer resolve a react this minor was never built against.
▎   - One exception: react-native and the four @react-native/* entries are written ^0.XX.0 — the minor floor, not the harvested patch. packages/plugin-verify-dependencies/__tests__/profile.ts requires that literal form, because a profile must accept any patch of its own minor.
▎   - A tracked key that disagrees with the template is a bug, regardless of which minor introduced it. Inheritance carries stale pins forward silently, and some predate the minor you are adopting. Fix it in your minor and say so in the PR, or file it — do not propagate it.
▎   - Do not add tracking for packages prior profiles did not track (ecosystem-native libs like screens, gesture-handler, reanimated are not profile-tracked yet; see step 5).
▎   - Earlier profiles are not the spec. The template is.
```

AGENTS.md
```diff
@@ -0,0 +1,32 @@
...
+## Working conventions
+
+- Open pull requests against `develop`, not `main` (releases cut from `main` via Changesets).
+- Conventional Commits (`type(scope): subject`), one logical change per commit.
+- Pull requests open as drafts; mark ready only when the gates applicable to the change are green.
+- Match existing conventions: profile file shape (`0.83.ts`), workflow style in `.github/workflows/`, commit scopes from recent history.
+
+## Skills
+
+Task-specific procedures live in `SKILL.md` files under `.agents/skills/` (the cross-client convention, so any compliant agent picks them up). Read the relevant one in full before starting the task it covers.
```
NickBurkhartBB: It looks like this doesn't actually auto pickup so not a convention yet. Could possibly symlink each .claude,.cursor,.gemini to make work, or maybe just drop the statement as it will find it given this file.

NickBurkhartBB (changes requested):
@deemaabdallah looks good, ran into one issue with versions running for 0.84, and some minor nitpicks.

.agents/skills/upgrade-rn-version/SKILL.md
```diff
@@ -0,0 +1,69 @@
...
+   - A tracked key that disagrees with the template is a bug, regardless of which minor introduced it. Inheritance carries stale pins forward silently, and some predate the minor you are adopting. Fix it in your minor and note it in the PR, or file it; do not propagate it.
+   - Do not add tracking for packages prior profiles did not track (ecosystem-native libs like screens, gesture-handler, reanimated are not profile-tracked yet; see step 5).
+   - Earlier profiles are not the spec. The template is.
+4. **Align and compile the example.** In `apps/example`: `npx flagship-code align-deps --profile 0.XX --fix`, then `yarn install`. Never hand-edit the example's React Native pins; the profile drives them. Then run the native build: `yarn workspace @brandingbrand/code-example prebuild --build internal --env prod --platform android --verbose`, and the iOS equivalent with `pod install`. A version is not adopted until it has compiled as a real app. The plugin test matrix passing alone is not sufficient.
```
NickBurkhartBB: we missed a scenario here. We also need it to align versions with the top level package.json. In testing RN 0.84, if we set example to react 19.2.3 but top level is ^19.2.0, we get two reacts and get runtime errors in example app

Also in test with Haiku, it just ran `npx flagship-code align-deps --profile 0.XX --fix`in top level folder instead of in the apps/example, probably better to use `yarn workspace @brandingbrand/code-example flagship-code align-deps --profile 0.84` so it doesn't miss as often.  It also didn't build iOS/Android, which it says is cause "I made an assumption tests passing = native build would work". so maybe the gate needs expanded just to clarify the steps and not just native builds above?

NickBurkhartBB (commented):
@deemaabdallah  this test was mostly successful in my tests with Opus/Haiku, just a couple details with versions again.

Also during one run it called out node versions against the CI builds in ./.github/workflows/pr-compile.yml, as node engine for 0.84 is >=22 but we have node 20 in the workflow, so maybe something to add later.

NickBurkhartBB (approved):

PR #2897, opened 2026-07-30
ci: notify a webhook when an issue is labeled ready-to-ship

deemaabdallah:
## Summary
- Adds a workflow that fires when an issue is labeled `ready-to-ship` and POSTs the issue number and title to a webhook stored in the `DISPATCH_WEBHOOK_URL` repo secret.
- Lets a downstream automation pick up an approved issue without anyone doing it by hand.

## Notes
- Uses `permissions: {}` (no repo write access).
- The issue title is passed via env var and `jq`, not interpolated into the shell, to avoid script injection.
- Requires the `DISPATCH_WEBHOOK_URL` repository secret to be set.
- Issue events fire workflows from the default branch, so this only runs once merged to `develop`.

### DIFF
--- a/.github/workflows/notify-ready-to-ship.yml
+++ b/.github/workflows/notify-ready-to-ship.yml
@@ -1,0 +1,29 @@
+name: Notify on ready-to-ship
+
+# When an issue is labeled `ready-to-ship`, POST its number and title to a
+# webhook (stored in the DISPATCH_WEBHOOK_URL secret) so a downstream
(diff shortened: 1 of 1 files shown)


NickBurkhartBB:
@deemaabdallah Could we leverage web hook from settings and move the label check to the backend? https://docs.github.com/en/webhooks/using-webhooks/creating-webhooks

NickBurkhartBB (approved):

PR #2901, opened 2026-08-03
docs: make the adoption handoff followable from the tracking issue

deemaabdallah:
The weekly release watcher files an `rn-adoption` tracking issue, and that issue is where an adoption starts. Three things on that path could not be followed as written.

## 1. The tracking issue did not name the adoption procedure

`buildIssueBody` linked only the maintainer guide. The guide is a maintainer walkthrough covering the whole surface, including creating a `packages/plugin-transform-template/src/transforms/<version>` directory and copying a new `packages/templates/supporting-files/<version>` directory. `.agents/skills/upgrade-rn-version/SKILL.md` designates both of those as stop-and-report conditions and puts those trees out of bounds for an adoption, because each is a tooling change rather than a version addition. Neither directory has been added since 0.82 and 0.77 respectively, and 0.83 was adopted without either.

So whoever opens the issue reaches the broader document first and the constrained procedure not at all. The issue body now leads with the skill, keeps the guide as background, and states that the skill governs where the two differ. The link is built from `GITHUB_REPOSITORY` so it stays correct for a fork.

## 2. Step 8 pointed at a file that is never present

Step 8 said to read the previous minor's changeset if it disagreed with the listed packages. Changesets are consumed when the release publishes, so `.changeset/` holds only unreleased ones. For 0.84 that means 0.83's adoption changeset is not there and never will be. This skill treats a step it cannot run as a blocker, and this one sits immediately before the PR.

The same fact survives in the published changelogs, so step 8 now names that check and records why `.changeset/` does not hold it:

```
grep -rl "add support for react-native 0.YY" packages/*/CHANGELOG.md
```

For 0.83 this returns `cli`, `cli-kit`, `plugin-verify-dependencies` and `templates`, matching the four packages step 8 lists.

## 3. Step 9 left the tracking issue open and asked for a reviewer that cannot be derived

`develop` is the default branch, so `Closes #<issue>` in an adoption PR closes its tracking issue on merge. Step 9 never asked for it, leaving the last step of the loop manual. It is now part of step 9 and the definition of done.

Step 9 also asked for review from the repository's code owners. There is no `CODEOWNERS` file and `require_code_owner_reviews` is off on `develop`, so that instruction can only be satisfied by guessing at a name. Reviewer assignment is now left to maintainers.

## Verification

- `node --test scripts/rn-release-watch.test.mjs`: 15/15 pass. The existing `buildIssueBody` test now asserts the skill link alongside the guide link it already checked.
- Dry run against live npm and the GitHub API, which is what the Thursday schedule will do: highest adopted `0.83`, pending `0.84, 0.85, 0.86`, all three matched to existing issues, nothing filed.
- Rendered body checked for a synthetic pending minor; the generated skill URL resolves (HTTP 200).
- No changeset: no published package changes. `scripts/` and `.agents/` are outside the workspaces, and the root is private.

### DIFF
--- a/.agents/skills/upgrade-rn-version/SKILL.md
+++ b/.agents/skills/upgrade-rn-version/SKILL.md
@@ -87,7 +87,15 @@ Never adopt more than one version in a PR, never stack one adoption on another's
    - `test` is a cached task and the variable is not part of its cache key, so a run following a plain `yarn test` can replay the earlier result. `--force` is what prevents that.
 
    The proof is the `test-templates (0.XX)` job on the pull request, which cannot exist unless the template directory does. The two native builds in step 5 are part of this gate and are not implied by these four commands.
-8. **Changeset.** One file, message `add support for react-native 0.XX`, minor bump on the packages an adoption touches: `code-plugin-verify-dependencies`, `code-templates`, `code-cli-kit`, `code-cli`. Read the previous minor's changeset rather than this list if the two disagree. Declare another package only if this adoption actually modified it; do not declare one because it seems related.
+8. **Changeset.** One file, message `add support for react-native 0.XX`, minor bump on the packages an adoption touches: `code-plugin-verify-dependencies`, `code-templates`, `code-cli-kit`, `code-cli`. Declare another package only if this adoption actually modified it; do not declare one because it seems related.
+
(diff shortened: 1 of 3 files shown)


deemaabdallah:
Superseded by #2902, which carries the same two fixes without the rest.

